PR #155, opened 2023-02-24
bugzilla wrapper functions using perceval

leilani-reich:
I would like to add three new functions to Kaiaulu.
1. download_bugzilla()
- Downloads Bugzilla data using either  Perceval's traditional or REST API Bugzilla backend depending on what value is passed for the "backend" parameter
2. parse_bugzilla()
- Parses Bugzilla data using Perceval's traditional Bugzilla backend. Comments are optional to parse.
3. parse_bugzillarest()
- Parses Bugzilla data using Perceval's REST API Bugzilla backend. Comments are optional to parse.


carlosparadis:
Thank you, I will get back to you on this. I suggest you start on the next part of the milestone since this may take me some time!

R/parser.R
```diff
@@ -4,6 +4,281 @@
...
+#' Parse Bugzilla data obtained from Perceval traditional Bugzilla backend
+#'
+#' @param perceval_path path to perceval binary
+#' @param bugzilla_json json object containing bugzilla issue data
+#' @param comments if true, the comments are parsed along with the issues
+#' @export
+#' @family parsers
+parse_bugzilla <- function(bugzilla_json,comments=FALSE){
+
+  json_issue_comments <- data.table(jsonlite::stream_in(textConnection(bugzilla_json),verbose = FALSE))
```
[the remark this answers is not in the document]
carlosparadis: I will retract the comment above. It seems at least for the retro API of bugzilla via perceval, the file is downloaded as a list of json files, where each line is a different json file. This will cause the `parse_json` function to fail (contrary to your Crawler, which worked for me). 

This being the case, I guess we are without much option than representing the list as a table of lists (which I find confusing to manipulate, as simply a list of lists).

R/parser.R
```diff
@@ -4,6 +4,281 @@
...
+  }
+
+  all_issues <- rbindlist(all_issues,fill=TRUE)
+  all_issues_comments <- rbindlist(all_issues_comments,fill=TRUE)
+
+  parsed_issues_comments <- list()
+  parsed_issues_comments[["issues"]] <- all_issues
+  parsed_issues_comments[["comments"]] <- all_issues_comments
+
+  return(parsed_issues_comments)
```
carlosparadis, line 129: I found the return of the function to be confusing. If comments flag is off, a list of 2 elements is returned, where the second is empty, but if the flag is on, then both tables are present. 

My suggestion is to simply return the tables themselves. If comments is off, return the data table of issues, if comments is on, then return a single table of issues + comments. I believe this is how Jira Issue behaves, no? It's been awhile.

carlosparadis (changes requested):

R/parser.R
leilani-reich: Hi Carlos, I think that parse_jira() returns the tables similarly to how I did it. The function docs for parse_jira() says the following:
https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L289-L290

And has a similar return:
https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L385-L389

carlosparadis: I see. So that's where you got this from. Unfortunately, `parse_jira` is not the best code to reference as a model beyond the nomenclature. This is because the downloader (which you are implementing) is not implemented in Kaiaulu: 

http://itm0.shidler.hawaii.edu/kaiaulu/articles/download_jira_data.html 

So there is less control over how the data is stored. That being said, the change I am requesting here can likely be applied to the function too. Could you make an issue for that and assign it to me?

leilani-reich: Alright, I will do that.

leilani-reich:
Hi, I made an update for the wrapper parser function that uses data from Perceval's traditional bugzilla backend, but I am still trying to figure out the parser function using Perceval's rest api backend. I hope to get that done soon. Thanks.

leilani-reich:
## bugzillarest crawler function updates

I have also updated the bugzillarest parser function with fixes. For this function, a lot of times I would get a warning "warning: item 25 has 0 rows but longest item has 1; filled with na" whenever I was making a field of a parse_issue datatable because sometimes that field could be null (ex. character(0)). I think the main field causing the issue was data.keywords, which is why I am using a condition to check the value and if it is null, set it to NA. I have tested this function on around 5 different bugzilla sites without error, which is why I've left data.keywords as the only field with the special check (also, this "ifelse" check I used I learned from https://stackoverflow.com/questions/65781329/equivalent-of-ternary-operator-in-r). I could check all of the fields for my datatable issues for safety, but I'm not sure if that would be bad for runtime (although ifelse is vectorized).

R/parser.R
carlosparadis: Were the returns fixed on the more recent PRs? Having a bit of a hard time assessing at this point since the PR grew big!

R/download.R
```diff
@@ -4,16 +4,24 @@
...
 #' @param perceval_path path to perceval binary
 #' @param bugzilla_site link to specific bugzilla site
 #' @param datetime fetch bugs updated since this date (in any ISO 8601 format, e.g., 'YYYY-MM-DD HH:mm:SS+|-HH:MM'))
 #' @param backend either traditional "bugzilla" or "bugzillarest" for Bugzilla >= 5.0 servers
+#' @param max_bugs the maximum number of bugs requested on the same query. This acts as the limit parameter
+#' in the Bugzilla REST API. Bugzilla sites may have specific limits set, so make sure to change the max_bugs
+#' parameter accordingly to correctly download the data when using the "bugzillarest" backend.
 #' @export
-download_bugzilla <- function(perceval_path, bugzilla_site, datetime, backend="bugzilla"){
+download_bugzilla <- function(perceval_path, bugzilla_site, datetime, backend="bugzilla", max_bugs=500){
```
leilani-reich, line 21: Malia was having issues using my download_bugzilla for Redhat Bugzilla using the bugzillarest backend, and I also noticed problems, so I tried it using max_bugs set to 20 and it worked. Thus, this is the same case as the bugzilla crawler issue with the pagination. I've added the "max_bugs" parameter which acts as the "limit" parameter so that users can adjust this accordingly.

carlosparadis: Just to be clear, when you say the same issue:

You are saying Perceval will defer to the user to make a best guess on the page limit, and if they don't have any idea what that is, it would skip from page 0-20 to 500-520, hence skipping issues on Red Hat case?

And, I see you made one change for the parameter here, but does that mean we also need to add that file parsing check to adjust the limit so it doesn't skip?

leilani-reich: Hi, yes that seems to be the case. When I change the max_bugs parameter, I noticed a different number of issues being returned.

>And, I see you made one change for the parameter here, but does that mean we also need to add that file parsing check to adjust the limit so it doesn't skip?

Do you mean saving each of the files from the Perceval calls into a folder and checking how many bugs are in the first one? Yes, I think we would need that check to adjust the limit. Should I save the data similar to how I did it for the crawler downloader then?

leilani-reich:
Addressing https://github.com/sailuh/kaiaulu/pull/179#issuecomment-1518977862 here as it seems more appropriate. I am specifically talking about the Perceval bugzilla data here.

>1. Now, on the issue request, does the API provides you with a issue description? I think this is where I'm being forgetful.

Just to be clear, I only made these changes for the data I got from Perceval for bugzilla, since I made the parser functions for these.

For download_bugzilla() using Perceval's traditional bugzilla backend,
I use the first comment as the issue description.


For download_bugzilla() using Perceval's bugzilla REST API backend,
there is a separate field for the description.

>Did you tell me that in the issue request, the issue_description has the first comment of the issue? I am asking explicitly what the API gives you, not what you coded to be clear.

For download_bugzilla() using Perceval's traditional bugzilla backend,
there is no explicit issue_description field, but the issue description is comment 0 within the long_desc field.

Screenshot 2023-04-22 at 10 45 53 PM

For download_bugzilla() using Perceval's bugzilla REST API backend,
there is a separate field for the description, and that description is the same as the 0th comment in the comments field in the json.

Separate description field (outside of comments field):
Screenshot 2023-04-22 at 10 47 00 PM

Description within comments field as comment with count 0:
Screenshot 2023-04-22 at 10 47 42 PM

carlosparadis:
In your comment above (not as a separate reply), could you edit so as to make it clear:

1.1) If Perceval downloads everything in one command, or if it distinguishes issue requests from comment requests. 
1.2) If it distinguishes, I need to know which of the two you are calling with Perceval on your examples above, the issue or the comments.

3) Also, could you add to your comment above what the RESTful API returns from your Crawler downloader for both issue and comments? I want to also know what Bugzilla API actually returns, not what Perceval gives you (which seems to be inconsistent depending on whether it is traditional or RESTful from what you are saying). 

I need to know this so I can answer @malialiu question on whether to make a separate column or not. We should stick closer to the raw data, that being what the API returns to a GET command, rather than what Perceval does on the downloader, and then correct Perceval for consistency in our parser function if it needs to be.

leilani-reich:
@carlosparadis I have updated my comment.

carlosparadis:
Whew! Okay, thank you so much for making this comprehensive post. 

I now also see and remember you asked me on our last meeting if, for Perceval's Bugzilla Restful, if we should do something about the duplicated comments, and I believe I told you to, on the Notebook overwrite one with NA, maybe. 

Okay, so here's what I suggest, and let me take the chance to also give you nomenclature suggestions for these functions while I have the full picture in my mind. 

1) First, we want to make sure what Kaiaulu output out of these 3 tables looks consistent both on column names and the values in them (to the extent possible). 

2) Moreover, since the Restful API gives us the raw data, we will abide by its expectations. Therefore, we should assume Bugzilla does not have a Description field (even if, in practice, developers use the first comment as one). Indeed, if anyone tries to look at your Notebook for example issues, you will see an issue summary (which is the title of the issue), but there is no visual distinction between issue description and issue comment (this is different from JIRA, which has a very distinct separation from issue description and comments). 

Since visually we can't tell them apart, and via the API the issue_description doesn't exist, then we should not artificially assume it either (as Perceval in one endpoint does inconsistent to itself). Concretely and using this as reference:

From comment: https://github.com/sailuh/kaiaulu/pull/155#issue-1598055427
```
download_bugzilla()
Downloads Bugzilla data using either Perceval's traditional or REST API Bugzilla backend depending on what value is passed for the "backend" parameter
parse_bugzilla()
Parses Bugzilla data using Perceval's traditional Bugzilla backend. Comments are optional to parse.
parse_bugzillarest()
Parses Bugzilla data using Perceval's REST API Bugzilla backend. Comments are optional to parse.
```

## Perceval Bugzilla Traditional 

### Function Nomenclature 

 * Remove the `backend` parameter here, and hardcode it to be the traditional backend on the system2 call. 

~~download_bugzilla~~ download_bugzilla_perceval_traditional_issue_comments()

~~parse_bugzilla()~~ parse_bugzilla_perceval_traditional_issue_comments()

### Parser Behavior

It seems the Perceval Traditional Downloader works consistent to the Bugzilla API, as it does not add a `issue_description` field. Do not add one either on your parser. Keep everything on the comments column. 


## Perceval Bugzilla RESTful

### Function Nomenclature 

 * Remove the `backend` parameter here, and hardcode it to be the Restful backend on the system2 call. 

~~download_bugzilla()~~ download_bugzilla_perceval_rest_issue_comments()

~~parse_bugzillarest()~~ parse_bugzilla_perceval_rest_issue_comments()

### Parser Behavior

Here Perceval is inconsistent with itself. Your parser should force it back into consistent: Delete the issue_description Perceval tries to add. Add on the description this behavior that Perceval is inconsistent to the (use \code\link to download_bugzilla_perceval_traditional_issue_comments()) other function. Also mention you delete the column due to redundancy to the comment column. With this, both the two Perceval functions should have the exact same output after your parser is done, but the downloader will differ. So also note on the download too that there is inconsistency and link the associated downloader.  

## Kaiaulu Perceval Downloader/Parser

### Function Nomenclature 

download_bugzilla_rest_issues()
download_bugzilla_rest_comments()
parse_bugzilla_rest_issues()
parse_bugzilla_rest_comments()

### Parser Behavior

@malialiu Finally, here you will abide by the same: Do not create a issue_description field on your parser function. Leave everything on comments.

@leilani-reich Your downloader should not modify the json, so I expect you did not try to add a issue_description field either. 

## Kaiaulu Bugzilla Notebook

In the notebook is where you should put your code that generates the issue description for all the 3 functions! Here, you can explain in text, alongside the example that, although Bugzilla API lacks a issue description field, you nonetheless observed that they are in essence a issue description field. Subsequently, add a line of code and create the column issue_description on the Notebook for all 3. 

This may appear redundant for the Perceval function, but it has a much more sane separation of concerns: Downloaders do not touch the data. Parsers sanitize the download to something that makes sense to the user and explain any issues (so they do not suffer the same headache we did trying to make sense of others code), then finally the Notebook discusses our own opinions and observations. 

## Misc

The function names may end up long, and the removal of the backend parameter may add a bit of redundant code, but having matching pairs of functions, and clarity of what I expect to receive when I run in their name is something I prefer over averaged sized functions I have to refer to the documentation every time. 

Let me know if I missed anything, but I hope this clarifies everything that is left to wrap this PR.

leilani-reich:
I'm so sorry, I completely missed this, but it seemed weird the issue didn't have a description when we get it from REST API. So I looked in https://bmo.readthedocs.io/en/latest/api/core/v1/bug.html#search-bugs and just noticed there is a section for adding extra fields, and description is one of them. Not to mention comments. So it would look like [this](https://bugzilla.redhat.com/rest/bug/35?include_fields=_default,description,comments).

Screenshot 2023-04-23 at 1 53 39 AM

carlosparadis:
So, let me see if I am getting this right using the specification I just posted on the comment`download_bugzilla_rest_issues()` could actually be `download_bugzilla_rest_issue_comm

leilani-reich:
Yes, that seems to be the case.... Here's an example using the creation_date https://bugzilla.redhat.com/rest/bug?creation_time=2023-04-23&include_fields=_default,description,comments&limit=20&offset=10.

carlosparadis:
Here's my suggestion since we are already strained on time:

Maintain the specification of https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1519036410. 

While it is reassuring Perceval group is staying true to the raw data, I still find it saner to not add a redundant issue description, especially when an issue description looks just like a comment on any Bugzilla issue. So, knowing the field "description" exists doesn't change anything for us.

However, knowing the comments field exist is...substantial. This would massively reduce the number of GET requests against the server for massive projects. This, I think, we should try to incorporate. But I would not modify the code of you already did, only apply the revisions of my specification above. I would add two other functions instead to Kaiaulu Downloader/Parser:
 
 * download_bugzilla_rest_issues_comments()
 * parse_bugzilla_rest_issues_comments()

Correct me if I am wrong, but modifying your download function into a new function should only require changing one word on your GET request. Is that correct?

As far as Malia's part is concerned, her existing  `parse_bugzilla_rest_comments()` contains all the code logic to parse a "comment" object within your issue object. So, if you borrow her code logic into a function after the JSON is loaded in memory, that should be the exact same as to what you need to parse the comment objects inside each issue. 

So the additional logic of the combined parser would be accessing the comments object when you iterate through the issues via Malia's existing function, plus combining the tables, as Perceval does. Is that right? 

I would shoot this for @malialiu last week revision on the parser, and keep your effort to the downloader only, since milestone 3.4 is far more important than these two additional functions at this point. 

Does that work?

carlosparadis:
> https://bugzilla.redhat.com/rest/bug?creation_time=2023-04-23&include_fields=_default,description,comments&limit=20&offset=10. 

To be clear, I would not include the "description" field on any of the requests. `default` i am not sure what it is doing, I am assuming it is what we already get. Then the only new field would be `comments`. Correct?

leilani-reich:
I need the _default so that we get the default fields and without it, I would just get comments (assuming I remove description). So yes, comments is the only new field for the json. I can add the new downloader. Sorry once again for overlooking this.

carlosparadis:
No worries. Just remember: Milestone 3.4 is still the priority! :)

malialiu:
> @malialiu Just to be clear, the function renaming is not yet implemented right?
> 
Sorry if I missed anything, so I should create a function```parse_bugzilla_rest_issues_comments <- function(bugzilla_issues, bugzilla_comments)``` to combine the data tables from ```parse_bugzilla_rest_issues()``` and ```parse_bugzilla_rest_comments()```?

carlosparadis:
Hi Malia, 

Your function should not be combining data. You should write a slightly modified version of Leilani's downloader: https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1519050276 

With the correct parameter, both issue and comment data will be retrieved on a single JSON. Your parser should then parse the files.

What I mentioned was that, the structure of the JSON is very similar to the one you already wrote your functions for. See: https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1519057617

carlosparadis:
@malialiu Have a look at this example Leilani used:

https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1519057056

Do you see how it has both issue and comment data in a single file? I expect your 


Screen Shot 2023-04-27 at 1 03 23 PM

`parse_bugzilla_rest_issues()`

To work on it. Could you try?

The part I expect your `parse_bugzilla_rest_issues()` to not capture is the `comments` field highlighted above. However, your `parse_bugzilla_rest_comments()` should contain the entire logic to parse the "comments" field. So really, your 

`parse_bugzilla_rest_issue_comments` should ideally be able to call both functions above and do very minimal work other than call the two functions in the appropriate fields.

leilani-reich:
reply to: https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1525499536

I need to do the function renaming on my own bugzilla pull requests. Also, I am doing the new downloader function, as we discussed [here](https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1519057617).

carlosparadis:
reply to https://github.com/sailuh/kaiaulu/pull/155#issuecomment-1526397325

@leilani-reich you sure you don't want to defer this to @malialiu? You already have several tasks for milestone 3.4, which takes priority over this.

leilani-reich:
Okay, I gave some guidance to Malia on how to change the downloader. I will work on finishing milestone 3.4. Actually, I am at the part of testing the generalization of the functions and am hoping to finish soon.

carlosparadis (approved):
Looks good and the checks are passing!

Issue #151, opened 2023-02-16
Improving Kaiaulu Documentation


leilani-reich:
We will add improvements to the Kaiaulu documentation. These will include clarity on the Social Smells notebook and other vignettes, grammar/spelling fixes, and improvements on the function docs.

Tasks:

Leilani:https://github.com/sailuh/kaiaulu/pull/158
Pull Request: 
- [x] Documentation fixes

Nico:
Pull Request: https://github.com/sailuh/kaiaulu/pull/161
- [x] Documentation fixes

Malia:
Pull Request: https://github.com/sailuh/kaiaulu/pull/157
- [x] Documentation fixes

carlosparadis:
Thank you! 

I would like to refine a bit what is being sought here. Several artifacts together compose the documentation of Kaiaulu: 

   1. (*) **README.md**
2. CONTRIBUTING.md (as pointed out in #150)
3. (*) config files (be them either the /conf and tools.yml). (I.e. consider the commentary around each field that may be lacking, for example #146, or the overall template itself) 
4. the wiki (as improved in #145)
5. the Kaiaulu major use cases (references.bib) 
6. (*) **The notebooks** (not all of them have a narrative explaining the process or the config file)
7. (*) **the documentation on each function in R/** (e.g. when you type `?parse_gitlog`. 
8. **pkgdown docs**: i.e. http://itm0.shidler.hawaii.edu/kaiaulu (this is auto generated by an R package named pkgdown, but it is still an easier way to navigate the function docs, vignettes and NEWS.md). 
9. the tags on GitHub used to annotate the issues, and the template of issues themselves (which is currently none)
10. Of course, the published paper itself gives the conceptual intent behind the tool and the vision, but that can't be edited. 
11. The man pagage of Kaiaulu CLI, which is the server-side endpoint of Kaiaulu (https://github.com/sailuh/kaiaulu/tree/master/exec) 

The (*) emphasizes the ones I believe are more critical, but if your usage of them suggests otherwise, then that is ok and helpful to know too!

For this deliverable, I would like you to take a comprehensive view of all the artifacts made available, which together form your user experience when first installing the package. Here's a suggestion of section template for this deliverable:

  1. Enumerate the documentation artifacts (as above, and/or any other I may have missed).
  2. The task you are trying to perform. This would be the one provided to you by @rnkazman with Apache Thirft. More specifically, using the package for computing Social Smells, composing the config file for Apache Thirft, and collaborating in GitHub (pull request). 
[...]

PR #158, opened 2023-02-26
improving kaiaulu documentation

leilani-reich:
I've made some changes to the documentation based on my comments in issue #151.


carlosparadis:
Thank you! I will get back to you on this too. On a minor note, you may want to revisit the commits for this one. `get updates from Kaiaulu master` should probably be combined before it is pushed into the main repository, but give me a chance to review them separately first before making further changes.

carlosparadis (changes requested):
This looks fine. Thank you for adding the mention to _pkgdown here. Could you git squash the "getting updates from kaiaulu" into one commit?

Also, please update the new commit message to be more specific. I.e. rather than "documentation improvements", mention instead "i #158 adds mention to _pkgdown.yml" instead or something of the sort. 

But otherwise 👍

leilani-reich:
Hi, I squashed the commits. Will this work? Thanks.

carlosparadis:
Screen Shot 2023-03-16 at 3 44 37 AM

I still see two commits, now because of my more recent, so you may need to squash again, I think? 

Also, when you do, try `git commit -s` and try making a shorter title and moving the details to the body. Otherwise, Github will truncate as you see above (doing title and body manually avoids the truncation if you can't make the title short enough).

leilani-reich:
Hi Carlos, it is all one commit now. I wasn't sure if I should have combined your commit as well, which is why I had it separate.

carlosparadis:
We forgot to include the .Rd files on the last commit. These are generated after you re-compile the Docs with any changes. I just added them on commit e3d3053eed82f6f65b638caa70168b6d7c255bad. We should include them on the other pull requests, including new functions (except for the unit tests since they don't have them).

PR #162, opened 2023-03-03
unit testing parser module

leilani-reich:
Here are my unit tests for parse_mbox and parse_jira.


carlosparadis (changes requested):
Several things need to be changed on these pull requests, so the unit tests serve their intended need, some of which I am not clear yet due to how much overhead you will need to generate the required input to the functions.

I think for now, start by just remaking the unit test that checks the perceval function and post to the Pull Request comments. I want to make sure we are on the same page in regards to loading the tools.yml for the user, and the assumption of path, as this applies to every unit test.

carlosparadis:
On project view, I have moved this PR to "backlog".

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
+test_that("Using Perceval version other than 0.12.24 will fail for parse_mbox", {
+  perceval_path <- "../../../tools/bin/perceval-incompatible-version"
```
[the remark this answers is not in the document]
leilani-reich: Hi Carlos,
does this look better?

```r
test_that("Perceval version is compatible", {
  tools_path <- "../../tools.yml"
  conf_path <- "../../conf/thrift.yml"
  tool <- yaml::read_yaml(tools_path)
  conf <- yaml::read_yaml(conf_path)
  perceval_path <-tool[["perceval"]]
  mbox_path <- paste("../", conf[["mailing_list"]][["mbox"]], sep="")
  expect_no_error(parse_mbox(perceval_path, mbox_path))
})
```
Also, just wanted to confirm, the "rawdata" folder we see in the config files is only one parent directory outside of Kaiaulu, right?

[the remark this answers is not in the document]
leilani-reich: Got it. I will change the hardcoding.

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
...
+  mbox_path <- "../../../../rawdata/mbox/project.mbox"
```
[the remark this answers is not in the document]
leilani-reich: Hi Carlos,
I split the test "Using Perceval version other than 0.12.24 will fail for parse_mbox" into two separate tests and fixed the hardcoding issue.

```r
test_that("Perceval version is 0.12.24", {
  tools_path <- "../../tools.yml"
  tool <- yaml::read_yaml(tools_path)
  perceval_path <-tool[["perceval"]]
  perceval_version <- system2(perceval_path, args="--version", stdout=TRUE, stderr=FALSE)
  expect_equal(perceval_version, "perceval 0.12.24")
})
```

```r
test_that("Perceval path works for parse mbox", {
  tools_path <- "../../tools.yml"
  conf_path <- "../../conf/thrift.yml"
  tool <- yaml::read_yaml(tools_path)
  conf <- yaml::read_yaml(conf_path)
  perceval_path <-tool[["perceval"]]
  mbox_path <- paste("../", conf[["mailing_list"]][["mbox"]], sep="")
  expect_no_error(parse_mbox(perceval_path, mbox_path))
})
```

How does this look?

Also, just wanted to confirm, the "rawdata" folder we see in the config files is only one parent directory outside of Kaiaulu, right?

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
...
+  expect_error(parse_mbox(perceval_path, mbox_path))
+})
+test_that("Correct config file parameters and Perceval version 0.12.24 work", {
+  perceval_path <- "../../../tools/bin/perceval"
```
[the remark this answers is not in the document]
leilani-reich: Hi Carlos,
I think I meant to put "../../../tools/perceval/bin/perceval". Since in the tools.yml, as the default it uses "~/perceval/bin/perceval", and perceval is usually stored in a bin folder from what I've seen. I will keep that in mind that we shouldn't have a bin folder within the tools folder.

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
+test_that("Using Perceval version other than 0.12.24 will fail for parse_mbox", {
+  perceval_path <- "../../../tools/bin/perceval-incompatible-version"
+  mbox_path <- "../../../../rawdata/mbox/project.mbox"
+  expect_error(parse_mbox(perceval_path, mbox_path))
+})
+test_that("Correct config file parameters and Perceval version 0.12.24 work", {
```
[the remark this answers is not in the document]
leilani-reich: Got it. I will split up my tests. I will have one to check the perceval version = 0.12.24, one to check both parameters of parse_mbox, and one to check the parse_jira parameter.

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
...
+  perceval_path <- "../../../tools/bin/perceval"
+  perceval_path_bad <- "/bad/perceval/path"
+  mbox_path <- "../../../../rawdata/mbox/project.mbox"
+  mbox_path_bad <- "/bad/mbox/path"
+  jira_issue_comments_path_bad <- "/bad/jira/path"
+  expect_error(parse_mbox(perceval_path_bad, mbox_path))
+  expect_error(parse_mbox(perceval_path, mbox_path_bad))
+  expect_warning(expect_error(parse_jira(jira_issue_comments_path_bad)))
+})
+test_that("Empty file path passed to parse functions fails", {
```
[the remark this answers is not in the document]
leilani-reich: Good point. I guess I was thinking of a specific case of an incorrect path. I will remove this test.

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
...
+  mbox_path_rel <- "../../../../rawdata/mbox/project.mbox"
+  perceval_path_abs <- "/Users/absolute-path/tools/bin/perceval"
+  mbox_path_abs <- "/Users/absolute-path/rawdata/mbox/project.mbox"
+  jira_path_abs <- "/Users/absolute-path/rawdata/issue_tracker/project_issue_comments.json"
+  expect_no_error(parse_mbox(perceval_path_rel, mbox_path_abs))
+  expect_no_error(parse_mbox(perceval_path_abs, mbox_path_rel))
+  expect_no_error(parse_mbox(perceval_path_abs, mbox_path_abs))
+  expect_no_error(parse_jira(jira_path_abs))
+})
+test_that("Parsers produce tables with correct columns and expected size", {
```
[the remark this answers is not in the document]
leilani-reich: Hi Carlos,
shouldn't we expect the user to have a configuration file set up as well with all the parameters for social_smells_showcase.Rmd? Also, do you have a preference for which config file the user is using in these tests (thrift, helix, geronimo, etc)?

Also, in regards to the mini git, could we add a test data file in the testthat directory that contains several commits from the kaiaulu git log and use that for the test? I was thinking of following this convention https://r-pkgs.org/testing-design.html#storing-test-data where there's a fixture folder with the test data.

tests/testthat/test-social-smells.R
```diff
@@ -0,0 +1,112 @@
...
+  expect_error(parse_mbox(perceval_path, mbox_path))
+})
+test_that("Correct config file parameters and Perceval version 0.12.24 work", {
+  perceval_path <- "../../../tools/bin/perceval"
+  mbox_path <- "../../../../rawdata/mbox/project.mbox"
+  jira_issue_comments_path <- "../../../../rawdata/issue_tracker/project_issue_comments.json"
+  expect_no_error(parse_mbox(perceval_path, mbox_path))
+  expect_no_error(parse_jira(jira_issue_comments_path))
+})
+test_that("Incorrect paths to files passed to parse functions fails", {
```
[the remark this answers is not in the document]
leilani-reich: Hi Carlos, I'm a little confused on this. 

If we were testing if "parse_mbox fails when the path to perceval binary is incorrect", wouldn't we want to make a hardcoded path to perceval that we know is incorrect instead of taking the user's path from tools.yml, which may or may not be correct?  

For testing something like "parse_mbox fails when the path to mbox binary is incorrect", I understand why we would take the user's tools.yml path to perceval. However, shouldn't the mbox path be manually set by us to be a bad path instead of reading it from a project config file?

tests/testthat/test-social-smells.R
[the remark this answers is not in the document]
carlosparadis: It should be one level above Kaiaulu, yes.

The Perceval test looks fine. 

The second test I am not sure what you are really testing. Try to refine it further, so that you can use expect_equal too.

tests/testthat/test-social-smells.R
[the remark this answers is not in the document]
carlosparadis: I see. Yes, let's just assume all the binaries are inside a tools folder rather than a tools/bin, since I am not sure what else the tools/ folder would be used for. Kaiaulu mostly relies on the third parties that can be self-contained in a single file (e.g. a .jar or a binary), to avoid cluttering and overhead.

tests/testthat/test-social-smells.R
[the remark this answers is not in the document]
carlosparadis: You may want to further refine the mbox part, as noted above, since I am a bit confused what it is the goal.

tests/testthat/test-social-smells.R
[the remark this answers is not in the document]
carlosparadis: > If we were testing if "parse_mbox fails when the path to perceval binary is incorrect", wouldn't we want to make a hardcoded path to perceval that we know is incorrect instead of taking the user's path from tools.yml, which may or may not be correct?

I think I got confused about what you were trying to test here, the parse_mbox() function, or the fact the user typed it wrong. If you are testing the function behavior to a wrong path, then you are correct you should hardcode.

tests/testthat/test-social-smells.R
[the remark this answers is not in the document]
carlosparadis: We can assume what is on Kaiaulu repo. For example, a config file we host. The tools.yml we have also has to be modified to comply to your new unit tests. We can then test if the user, in modifying the configs to their needs, are not violating the assumptions we set forth in the repo. But since we do not have the data hosted, we can't assume the data.

No preference for the choice of config file, so long it is already in the repo. 

I am not sure if we want to start hosting test data, because I am not confident this will scale. Kaiaulu is a tool that has interoperability to a lot of different data sources. It implements some features of its own, but at its most basic purpose, it gives a common model to reason and combines all of them. If we start hosting test data for every single interface, our repo will clog, and I am not sure the folks over CRAN will be too happy either whenever Kaiaulu is made into an official R package. 

I'd rather we create the sample data on the fly instead. This will also give us more control over what we wish to test. For example, if we extend the git log sample creation function, to parameterize the number of files, and number of commits, we can now create very precise unit tests to see how parse_gitlog() behave. Meanwhile, using any sample git log, like Kaiaulu's, we may be missing out on these scenarios. Likewise to different conventions of annotating issue id on commits, and the body, which are parsed to construct issue networks, etc. 

I think hosting sample data has its uses, for example, if we were testing a machine learning algorithm which has randomize initialization, off the same data source over and over again. But I may reconsider this in the future.

carlosparadis:
@leilani-reich sorry I missed your in-line replies. I think e-mail notifications are not generated to these. Please feel free to post a comment here once you reply in-line, so I don't miss them in the future!

NEWS.md
```diff
@@ -41,6 +40,7 @@ kaiaulu [0.0.0.9600](https://github.com/sailuh/kaiaulu/milestone/5) (in developm
 ### DOCUMENTATION FIXES
 
  * New configuration file for the Apache Thrift project called "thrift.yml". [#148](https://github.com/sailuh/kaiaulu/issues/148)
+ * Adds unit tests for parse_mbox() and parse_jira(). [#154](https://github.com/sailuh/kaiaulu/issues/154)
```
nicoelee123, line 43: Leilani informed me that parse_mbox would no longer be tested since we'd need .mbox file. I can delete these tests if need.

tests/testthat/test-parser.R
```diff
@@ -1,52 +1,54 @@
...
-  expect_no_error(parse_mbox(perceval_path_abs, mbox_path_abs))
-  expect_no_error(parse_jira(jira_path_abs))
+test_that("Incorrect perceval path fails parse_mbox", {
+  conf_path <-"../../conf/thrift.yml"
+  conf <- yaml::read_yaml(conf_path)
+  mbox_path <- conf[["mailing_list"]][["mbox"]]
+  incorrect_perceval_path <- "/incorrect/path/to/perceval"
+  expect_error(parse_mbox(incorrect_perceval_path, mbox_path), "error in running command")
+})
+test_that("Incorrect mbox path fails parse_mbox", {
```
nicoelee123, line 33: I realize we might not be testing parse_mbox anymore, but an error that I had with this was that I was checking if we'd encounter a precise error if an incorrect path to .mbox file is passed. However, I copy and pasted the exact error message to check, but the test still fails after giving that exact error. The tests above and below are designed the same way and they passed. Again, I can delete this altogether since it involves parse_mbox

tests/testthat/test-parser.R
```diff
@@ -1,52 +1,54 @@
...
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
+  incorrect_mbox_path <- "/incorrect/path/to/mbox"
+  expect_error(parse_mbox(perceval_path, incorrect_mbox_path), "Items of 'old' not found in column names: [data.body]. Consider skip_absent=TRUE.")
+})
+test_that("Incorrect jira issue comments path fails parse_jira", {
+  incorrect_jira_issue_comments_path <- "/incorrect/path/to/jira_issue_comments"
+  expect_error(parse_jira(incorrect_jira_issue_comments_path), "cannot open the connection")
 })
 test_that("Parsers produce tables with correct columns and expected size", {
```
nicoelee123, line 44: I read the review about creating a mini .git file on the fly so that the test remains contained. I'm currently trying to figure that out. I can also delete all references to parse_mbox in this test as well.

carlosparadis: Hi Nico, 

The function already exists. 

 * http://itm0.shidler.hawaii.edu/kaiaulu/reference/git_create_sample_log.html 

 * http://itm0.shidler.hawaii.edu/kaiaulu/reference/git_delete_sample_log.html

NEWS.md
carlosparadis: The parse_mbox() tests should not be deleted. You should use their logic on the parse_gitlog() combined to the function that creates a sample git log that I pointed out in the other comment.

tests/testthat/test-parser.R
nicoelee123: I see the function, but I guess I'm just sort of confused. The function creates a sample .git on the fly, but in order to test parse_jira (and no longer parse_mbox), wouldn't I also need the sample repo to have a Jira project key?

carlosparadis:
Make sure to do one more merge from the main repo to this PR too. I see it is missing the more recent action for test-coverage.

NEWS.md
nicoelee123: @carlosparadis Do you mean we would need to try and create a .mbox file on the fly and then delete it after the test?

carlosparadis: No. Instead of unit testing mbox, unit test the parse_gitlog() which we already have a function to create it on the fly. Say, if you are testing what happens if the path to a .mbox file passed doesn't exist in parse_mbox(), then transfer the logic to the parse_gitlog() by unit testing what happens if the path to the .git file is wrong. Same logic, different functions they are applied to.

nicoelee123: I see, so get rid of the parse_mbox tests and replace it with parse_gitlog tests, but using the same logic as the previous parse_mbox tests.

carlosparadis: Correct! If the logic of the unit tests is preserved by transferring to parse_gitlog(), I can always replicate it back down to parse_mbox() down the road when the generate on the fly tests for mbox are created. They are fundamentally parser functions at the end of the day.

nicoelee123:
image
It seems like this is the only issue and I'm not sure how to fix this?

carlosparadis:
@nicoelee123 I am not sure what is going on either. For now, can you execute the tests locally and see if all pass? 

RStudio -> Build Tab -> More -> Test Package. 

Also, can you give me some idea of where you are at with the unit tests of milestones 1 and 2? Thanks!

nicoelee123:
The unit tests in the most recent commit passed using Build Tab -> Test Package in both this PR and Malia's PR

Milestone 1
- Completed Leilani's unit tests for parse_jira and parse_mbox
    - Made changes according to comments
    - Got rid of unit tests relying on existing data but saved the logic for parse_gitlog()
- Completed Malia's unit tests for git_checkout() and parse_gitlog()
    - Added parse_gitlog() tests based on logic from Leilani's parse_mbox tests

Milestone 2
- Still need to complete Bugzilla crawler unit tests
- Still need to complete Bugzilla wrapper function unit tests

carlosparadis:
@nicoelee123 Hi Nico,

Just to be clear are you saying this and Malia's PR is ready to be merged?

Also, could you run another merge from the main branch to the Pull Requests? That should fix the problematic checks so I can see them passing here too. Thanks!

carlosparadis:
@nicoelee123 

Hi Nico,

>   [ FAIL 6 | WARN 7 | SKIP 1 | PASS 32 ]


The tests are failing. Have a look at the checklog to see if you can diagnose.

nicoelee123:
image
This is the result after running the unit tests for this PR. I'm not sure why the tests are failing during the checks.

carlosparadis:
You missed my other question here: 

https://github.com/sailuh/kaiaulu/pull/162#issuecomment-1528878520

nicoelee123:
Yes, the tests are ready to be merged, just trying to find solution to the errors on the tests you have in place

carlosparadis:
The problem may be due to the fact the remote machine does not have Perceval. I will try running them local to my machine to see if it resolves.

tests/testthat/test-parser.R
```diff
@@ -0,0 +1,44 @@
+test_that("Using Perceval version 0.12.24", {
+  tools_path <- "../../tools.yml"
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
```
carlosparadis, line 4: `perceval_path <- path.expand(perceval_path)`
Needs to be added after this line, or it will cause the test to fail (~/ paths will not resolve without this).

tests/testthat/test-parser.R
```diff
@@ -0,0 +1,44 @@
+test_that("Using Perceval version 0.12.24", {
```
carlosparadis, line 1: "Perceval version 0.12.24 is being used"

tests/testthat/test-parser.R
```diff
@@ -0,0 +1,44 @@
...
+  tools_path <- "../../tools.yml"
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
+  perceval_version <- system2(perceval_path, args="--version", stdout=TRUE, stderr=FALSE)
+  expect_equal(perceval_version, "perceval 0.12.24")
+})
+test_that("Correct perceval path", {
```
carlosparadis, line 8: "Perceval path in tools.yml is specified correctly"

tests/testthat/test-parser.R
```diff
@@ -0,0 +1,44 @@
...
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
+  expect_equal(file.exists(perceval_path), TRUE)
+})
+test_that("Correct mbox path", {
+  conf_path <-"../../conf/thrift.yml"
+  conf <- yaml::read_yaml(conf_path)
+  mbox_path <- conf[["mailing_list"]][["mbox"]]
+  expect_equal(file.exists(mbox_path), TRUE)
+})
```
carlosparadis, line 19: What this is really testing is that the data is available. This test will always fail. We can only test for what is specified on the repo. The third party tools is the borderline case of this type of unit tests. In that case, it is fine the test fails as this serves as a reminder there are third party tools the user has to download, on a specific version, and ideally in a specific place. User can then try to follow the README, then re-run unit tests to see if they pass. 

This makes sense and saves us time, because there are only a few tools needed. For data, not so much.

carlosparadis: Let's use this instead:

```r
test_that("configuration files are placed on recommended path ", {
  conf_path <-"../../conf/thrift.yml"
  expect_equal(file.exists(conf_path), TRUE)
})
```

tests/testthat/test-parser.R
```diff
@@ -0,0 +1,44 @@
...
+test_that("Correct jira issues comments path", {
+  conf_path <-"../../conf/thrift.yml"
+  conf <- yaml::read_yaml(conf_path)
+  jira_issue_comments_path <- conf[["issue_tracker"]][["jira"]][["issue_comments"]]
+  expect_equal(file.exists(jira_issue_comments_path), TRUE)
+})
```
carlosparadis, line 25: Since this is another "does data exist" test, and we already test for the conf file, let's remove it.

tests/testthat/test-parser.R
```diff
@@ -0,0 +1,44 @@
...
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
+  incorrect_mbox_path <- "/incorrect/path/to/mbox"
+  output <- parse_mbox(perceval_path, incorrect_mbox_path)
+  expect_equal(nrow(output), 0)
+})
+test_that("Incorrect jira issue comments path fails parse_jira", {
+  incorrect_jira_issue_comments_path <- "/incorrect/path/to/jira_issue_comments"
+  expect_error(parse_jira(incorrect_jira_issue_comments_path), "cannot open the connection")
+})
```
carlosparadis, line 44: This test is generating a warning, instead of just a PASS. You may need to use tryCatch or some other mechanism to prevent the warning from leaking. 

```r
Warning (test-parser.R:36:3): Incorrect jira issue comments path fails parse_jira
cannot open file '/incorrect/path/to/jira_issue_comments': No such file or directory
Backtrace:
  1. testthat::expect_error(...)
       at test-parser.R:36:2
  6. kaiaulu::parse_jira(incorrect_jira_issue_comments_path)
  7. jsonlite::read_json(json_path)
       at kaiaulu/R/parser.R:323:2
  8. jsonlite::parse_json(...)
  9. jsonlite:::parse_and_simplify(...)
 10. jsonlite:::parseJSON(txt, bigint_as_char)
 11. jsonlite:::parse_con(txt, bigint_as_char)
 13. base::open.connection(con, "rb")
 14. ```

[ FAIL 0 | WARN 1 | SKIP 1 | PASS 37 ] incorrect_jira_issue_comments_path <- "/incorrect/path/to/jira_issue_comments"

carlosparadis (changes requested):
@nicoelee123 

Here are some more specific change requests. I asked in the past that the test descriptions are reworded properly. The latter unit tests without comments are fine. But the former need some editing. Please make sure the other Pull Request has test descriptions that conform to these. 

One unit test should be removed, other modified, etc.

Finally there is a warning being generated that needs to be addressed.

Can you give me an OK on the other pull request that it accounts for the review recommendations here before I try them and link to this comment on it?

Let me know if you have any questions. 

Thanks.

carlosparadis (approved):
Ok, this is passing locally for me too. I will try to diagnose the remote myself down the road since this is secondary.

Issue #168, opened 2023-03-20
DV8 Integration Functions - Leilani


leilani-reich:
This is a followup to issue #165, specific to my tasks for DV8 Integration.

TASKS:

Pull Request: https://github.com/sailuh/kaiaulu/pull/171

Milestone 3.3:
- [x] dv8_clsxb_to_clsxj() (core:export-cluster)
- [x] parse_dependencies() changes ~dv8_depends_to_sdsmj()~ (call Depends command for sdsm.json)
- [x] parse_dv8_clusters()

Milestone 3.4:
- [x] dependencies_to_sdsmj()
- [ ] gitlog_to_hdsmj()

leilani-reich:
Hi Carlos,
I have made wrapper functions for dv8_clsxb_to_clsxj() and dv8_depends_to_sdsmj().

## dv8_clsxb_to_clsxj() 

### core:export-cluster command Info

```
usage: dv8-console core:export-cluster [-h] [-outputFile <OUTPUTFILE>] [-xml] INPUT_FILE
Export clustering as a separate XML or JSON file
 -h,--help                  Print usage information for a command
 -outputFile <OUTPUTFILE>   The output file to create
 -xml                       Export in XML (instead of JSON) format
```

### Wrapper function

```r
#' Exports DV8 clustering (.dv8-clsx) as a separate JSON file
#'
#' @param dv8_path path to dv8 binary
#' @param clustering_path path to clsx to convert to json
#' @param output_path path to save output file
#' @export
dv8_clsxb_to_clsxj <- function(dv8_path, clustering_path, output_path){
  system2(dv8_path, args=c('core:export-cluster', '-outputFile', output_path, clustering_path), stdout=FALSE, stderr=FALSE)
}
```

Note: I did not include the "-xml" option since you had mentioned we only cared about json output.

### Example of calling wrapper function

```r
# Load in the DV8 file
tools_path <- "../tools.yml"
tool <- yaml::read_yaml(tools_path)
dv8_path <-tool[["dv8"]]

# Load in the config file
conf_path <- "../conf/helix.yml"
conf <- yaml::read_yaml(conf_path)

clustering_path <- conf[["analysis"]][["dv8"]][["clustering"]][["clsxb"]]
output_path <- conf[["analysis"]][["dv8"]][["clustering"]][["clsxj"]]
dv8_clsxb_to_clsxj(dv8_path, clustering_path, output_path)
```

## dv8_depends_to_sdsmj()


[...]

Issue #165, opened 2023-03-08
DV8 Integration Milestone Issue + Cheat Sheet


leilani-reich:
We will integrate the low-level function of the tool DV8 by [Archdia](https://archdia.com/) into Kaiaulu. 

To do so, we will write wrapper functions for DV8's command line interface (CLI) functions.
These functions involve locating anti-patterns, constructing design structure matrices (DSM), detecting hotspots, analyzing roots, quantifying debt, analyzing return on investment analysis, and more.

We will focus on a subset of functions for Kaiaulu, especially ones that can be chained together.
In particular, we will focus on creating wrapper functions for architecture flaws and the decoupling level metric.

For each function, we will create a unit test using the testthat package for R.

Finally, we will create at least two R notebooks (one for architecture flaws, one for the decoupling level metric) to showcase the wrapper functions and show step-by-step how a user would conduct a software analysis using these functions.

TASKS:

Pull Request: https://github.com/sailuh/kaiaulu/pull/171

Leilani Reich
Milestone 3.3:
- [x] dv8_clsxb_to_clsxj() (core:export-cluster)
- [x] parse_dependencies() changes ~dv8_depends_to_sdsmj()~ (call Depends command for sdsm.json)
- [x] parse_dv8_clusters()

Milestone 3.4:
- [x] dependencies_to_sdsmj()
- [ ] gitlog_to_hdsmj()

Nicholas Lee
Milestone 3.3:
- [x] dv8_gitlog_to_gitnumstat() (performs gitlog –numstat –date=iso)
- [x] dv8_gitnumstat_to_hdsmb() (scm:history:gittxt:convert-matrix)
- [x] dv8_hsdsmb_to_decoupling_level() (metrics:decoupling-level)
- [x] dv8_hsdsmb_to_hierclsxb() (dr-hier:dr-hier)
- [x] dv8_hsdsmb_drhier_to_excel() (office:export-spreadsheet)
- [x] dv8_dsmj_to_dsmb() (core:export-matrix)
- [x] parse_dv8_metrics_decoupling_level()

Malia Liu
Milestone 3.3:
- [x] dv8_hsdsmb_to_flaws() (arch-issue:arch-issue)
- [x] dv8_dsmj_to_dsmb() (core:convert-matrix)
- [x] parse_dv8_architectural_flaws()

Milestone 3.4:
- [x] dv8_hsdsmb_sdsmb_to_hsdsmb() (core:merge-matrix)

Milestone 3.5:
[...]

PR #171, opened 2023-03-26
dv8 functions leilani

leilani-reich:
Here are my current implemented functions for dv8 integration. I've added dv8_clsxb_to_clsxj(), parse_dv8_clusters(), and updates to parse_dependencies().

I've created a new file "dv8.R" to hold the functions, as discussed in https://github.com/sailuh/kaiaulu/issues/165#issuecomment-1469074768.

In regards to https://github.com/sailuh/kaiaulu/issues/168#issuecomment-1483783147, I've checked the parser.R file functions and parse_gitlog() calls the file column "file_pathname" https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L80, so I changed my file column name to this. However, I also saw the file column called "filename" in parse_git_blame() https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L174, and "filepath" in parse_line_type_file() https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L717.

I will later update parse_dv8_clusters() to use vectorization instead of for loops.

I will git squash all commits into one at the end.


carlosparadis:
@leilani-reich Thanks for catching the inconsistency in function names! Let's go with `filepath`. Could you make one more issue and copy and paste your observations on naming inconsistencies? 

Changing them may affect other Notebooks, so I need to look them more carefully before I standardize to filepath. You can call the issue name "Standardize filepath column name across tables".

Thanks!

R/dv8.R
```diff
@@ -0,0 +1,73 @@
...
+#' @param dv8_path path to dv8 binary
+#' @param clustering_path path to clsx to convert
+#' @param output_path path to save output file
+#' @export
+dv8_clsxb_to_clsxj <- function(dv8_path, clustering_path, output_path){
+  system2(dv8_path, args=c('core:export-cluster', '-outputFile', output_path, clustering_path), stdout=FALSE, stderr=FALSE)
+}
+#' Parse a clustering json from DV8
+#'
+#' @param clustering_path path to clustering json (obtained from dv8_clsxb_to_clsxj)
```
carlosparadis, line 18: There is a specific syntax you can use here. Search the git repo for the syntax:

#' @seealso \code{\link{parse_gitlog}}

And add below the @family parsers line. This will significantly facilitate code navigation of the pipeline via the R docs function, since they hyperlink on R Studio. Please apply it as needed to the other function docs.

R/dv8.R
```diff
@@ -0,0 +1,73 @@
...
+#' Exports DV8 clustering (.dv8-clsx) as a separate JSON file
+#'
+#' @param dv8_path path to dv8 binary
+#' @param clustering_path path to clsx to convert
+#' @param output_path path to save output file
+#' @export
+dv8_clsxb_to_clsxj <- function(dv8_path, clustering_path, output_path){
+  system2(dv8_path, args=c('core:export-cluster', '-outputFile', output_path, clustering_path), stdout=FALSE, stderr=FALSE)
+}
+#' Parse a clustering json from DV8
```
carlosparadis, line 16: Parse a DV8 json cluster file to R data.table.

R/dv8.R
```diff
@@ -0,0 +1,73 @@
...
+        filename_k <- structure[[i]]$nested[[j]]$nested[[k]]$name
+
+        # Represents a row in the table with the layer #, module #, and filename
+        row_ijk <- list(filename_k, module_j, layer_i)
+
+        # Append current row (made up of the layer number, module number, & filename) to a list
+        cluster_list <- append(cluster_list, list(row_ijk))
+      }
+    }
+  }
```
carlosparadis, line 63: Try to remove the for loop and vectorize the code.

R/parser.R
```diff
@@ -539,10 +539,11 @@ parse_github_replies <- function(github_replies_folder_path){
 #'
 #' @param depends_jar_path path to depends jar
 #' @param git_repo_path path to git repo (ends in .git)
+#' @param output_dir path to output directory (formatted output_path/)
 #' @param language the language of the .git repo (accepts cpp, java, ruby, python, pom)
 #' @export
 #' @family parsers
-parse_dependencies <- function(depends_jar_path,git_repo_path,language){
+parse_dependencies <- function(depends_jar_path,git_repo_path,language,output_dir="/tmp/"){
```
carlosparadis, line 546: ~~I am not sure why this was added.~~

I believe I asked you to parameterize tmp. Could you refer me to the comment I asked this so I remember why?

carlosparadis (changes requested):
Thank you for sending an early pull request! :) 

I have added a few notes on documentation formatting and variable nomenclature. This is more a matter of personal taste, which is why I wanted to iterate as we go (so you could use as reference for the next functions). 

I will share this PR review with your group, but please bring to their attention just in case, so I do not need to mention to them too (especially the \code\link one!). 

Feel free to add more commits before reviewing, or vice-versa. But I will assume the following commits will implement the requested changes, or that otherwise it was an oversight and I will point out again.

R/parser.R
leilani-reich: Hi Carlos, here's the relevant comment:
https://github.com/sailuh/kaiaulu/issues/168#issuecomment-1478886665

carlosparadis: :) Thank you! Yes, this is perfect.

carlosparadis:
Hi, 

A minor addition I noticed based on the code review to Malia: 

The function docs can have both the first line explaining it, but also a small section below for the description. Please include on the body of function docs, the naming of the file that will be generated, as noted here:

https://github.com/sailuh/kaiaulu/pull/174#discussion_r1154245373

This will help new users understand what we mean by a json or binary file in dv8. 

Screen Shot 2020-06-09 at 1 10 38 AM

Namely, the body of every function should clarify what we mean by a binary file according to dv8 nomenclature (as noted in the comment above to Malia). Which means a brief remark to the file format in dv8. 

We should probably ensure our own nomenclature for the binary and json files is consistent to the image above too. I may have made a mistake on the pipeline diagram for the file names, but we can revisit these after since it is just a small change on the string of how we save the files.

leilani-reich:
Hi Carlos, I will add the description of the file format to my function docs. Also, I think there's a mistake in that picture. It says *-sdsm.dv8-clsx, *-hdsm.dv8-clsx, and *-merged.dv8-clsx, but shouldn't these be dv8-dsm files instead of dv8-clsx files?

leilani-reich:
Hi Carlos,
I have added a commit for the dependencies_to_sdsmj() function and added a dv8 notebook to show how I called the functions. This is a work in progress for now, and I will later fix the hardcoded paths in my notebook, and vectorize the code. Also, I believe I will need to sort my json data by the smallest src to greatest src, which I will revisit.

I had some concerns regarding the dependencies_to_sdsmj() function:

1. What should the @schemaVersion be? This is one of the fields used in the sdsm.json (attached), but I am not using a schema. 

2. Should I let the user type the name of the project? I added this as a parameter because there is a "name" field in the sdsm.json with the name of the project, and I'm not sure how to get this otherwise.

3. On the note of the "name" field, I tried to use auto_unbox to get it to be just a string value, but the json still generates the name values in a one-element list (see attached json called depends_to_sdsmj_result.json).


4. Do my parameters need to be in the exact order used by dv8?

In the sdsm.json, it looks like this:
```
    {
      "src": 421,
      "dest": 64,
      "values": {
        "Call": 2,
        "Import": 1,
        "Use": 5,
        "Parameter": 3,
        "Contain": 1,
        "Return": 2
      }
```
Do I need to have my "dest", "values", "Call", etc written in the same order as DV8 formats it?

5. My biggest concern is the variables field in the sdsm.json. In particular, I thought this variable field would be a list composed of all the unique src and dest filenames. However, when I made a list of all the unique src and dest filenames from my parsed depends table, it was of smaller length than the variables array in the sdsm.json, by around 20-30 in size (1416 vs 1390). I'm still not sure what the difference is here.

I have attached an example of the sdsm json (Depends-Tests-0.96a-sdsm.json) and the json I'm currently getting from my dependencies_to_sdsmj() function (depends_to_sdsmj_result.json).

R/dv8.R
leilani-reich: I had the @family parsers line before I believe.

carlosparadis:
> 1. What should the @schemaVersion be? This is one of the fields used in the sdsm.json (attached), but I am not using a schema.

I am not sure about this. Could you e-mail Yi about this and Cc Rick, I, and Yuanfang? Do you have their e-mails? Otherwise I can send it to you. 

> 2. Should I let the user type the name of the project? I added this as a parameter because there is a "name" field in the sdsm.json with the name of the project, and I'm not sure how to get this otherwise.

Hmm..

On this issue:

https://github.com/sailuh/kaiaulu/issues/168#issuecomment-1475972460

```
tool:
  # Depends allow to parse file-file static dependencies.
   dv8:
      # the project folder path files in function calls will be used. 
      # The folder name will be used in the file names.
      folder_path: ../../rawdata/dv8_thrift
      # the architectural flaws thresholds that should be used
      architectural_flaws:
         clique_depends: call,use
         crossing_cochange: 2
         .
         .
         .
         change_cover: ?            
```

We assume the user must specify a folder_path where all the data gets stored. Let's take that folder path containing the name of the project. 

That being said, I think this assumption should be encoded on the R Notebook. I.e. the code to parse and extract the last folder in the path lives on the R Notebook. The path of the output file can then be constructed also on the notebook. Regarding the functions, I guess we can keep them simple and close to the commands from DV8. You expect an input path and output path variables where applicable. The R Notebook does the string concatenation for consistency when passing the parameters.

Does that make sense? That keeps the function more flexible if the user wants to use it for something else, and the R Notebook serves as the facilitator relying on the project configuration file. 

> 3. On the note of the "name" field, I tried to use auto_unbox to get it to be just a string value, but the json still generates the name values in a one-element list (see attached json called depends_to_sdsmj_result.json). Any suggestions how to fix this?

I am not sure what the auto_unbox function is. I believe you want to represent your json data in R as a list, since I'd expect jsonlite write function to map it 1 to 1 an R list to a json. If this doesn't answer the question can you give me an example of the issue or elaborate?

> 4. Do my parameters need to be in the exact order used by dv8?

This would be a question for Yi. I do not know how the source code of DV8 assumes the input, but I'd recommend staying consistent to the example files so any change in the future where DV8 cares about this, wouldn't break your function.

> 5. My biggest concern is the variables field in the sdsm.json. In particular, I thought this variable field would be a list composed of all the unique src and dest filenames. However, when I made a list of all the unique src and dest filenames from my parsed depends table, it was of smaller length than the variables array in the sdsm.json, by around 20-30 in size (1390 vs 1416). I'm still not sure what the difference is here.

Read both yours and the mismatching file variables as a character vector in R. Then use the function setdiff to identify the 20-30 that are unique. 


> EDIT: I have also made the changes you requested previously aside from vectorizing parse_dv8_clusters. I would like to return to that after finishing my milestone 3.4 functions.

I will take a look at your new commit. Thanks!

Keep in mind there are a number of changes that I am also going by guesswork here until I see the first full version of them. So we may need to revisit some assumptions I am proposing here after I see the first version, but I expect them to be mostly minimal like nomenclature.

Please point out if I missed anything.

R/dv8.R
```diff
@@ -4,28 +4,28 @@
...
+#' Transforms a DV8 binary cluster file to a json cluster file.
 #'
 #' Converts a clustering binary in the dv8 format *-clsx.dv8-cslx into a
 #'  *-clsx.json.
 #'
 #' @param dv8_path path to dv8 binary
-#' @param clustering_path path to clsx to convert
-#' @param output_path path to save output file
+#' @param clsxb_path path to clsx to convert
+#' @param clsxj_path path to save output file
```
carlosparadis, line 14: Yes this looks good!

R/dv8.R
```diff
@@ -4,28 +4,28 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-#' Exports DV8 clustering as a separate JSON file
+#' Transforms a DV8 binary cluster file to a json cluster file.
```
carlosparadis, line 7: Ok!

R/dv8.R
```diff
@@ -4,28 +4,28 @@
...
 #'
 #' Converts a clustering binary in the dv8 format *-clsx.dv8-cslx into a
 #'  *-clsx.json.
```
carlosparadis: I see you actually already included the dv8 format of the files, so it is clear to the user what we are calling a binary file. Ok!

R/dv8.R
```diff
@@ -4,28 +4,28 @@
...
+#' @param clsxb_path path to clsx to convert
+#' @param clsxj_path path to save output file
 #' @export
-dv8_clsxb_to_clsxj <- function(dv8_path, clustering_path, output_path){
-  system2(dv8_path, args=c('core:export-cluster', '-outputFile', output_path, clustering_path), stdout=FALSE, stderr=FALSE)
+dv8_clsxb_to_clsxj <- function(dv8_path, clsxb_path, clsxj_path){
+  system2(dv8_path, args=c('core:export-cluster', '-outputFile', clsxj_path, clsxb_path), stdout=FALSE, stderr=FALSE)
 }
-#' Parse a clustering json from DV8
+#' Parse a DV8 json cluster file to R data.table.
```
carlosparadis, line 19: Great!

carlosparadis (approved):
These look good! I will hold on merging the PR until we iterate over the next commits too. Then we can check for consistency across all the group docs to make sure the docs looks the same for the DV8 module.

leilani-reich:
> > 1. What should the @schemaVersion be? This is one of the fields used in the sdsm.json (attached), but I am not using a schema.
> 
> I am not sure about this. Could you e-mail Yi about this and Cc Rick, I, and Yuanfang? Do you have their e-mails? Otherwise I can send it to you.

Hi Carlos, I have their emails. Actually, I was looking at the Depends repo and they show how the output json should look https://github.com/multilang-depends/depends/blob/master/doc/output_format.md. I have only seen "schemaVersion 1.0" used, but I will email about this to check.

In addition, I wanted to make sure, do you want the sdsm.json output by the dependencies_to_sdsmj()  to look exactly like the Depends command output json, besides whatever the user filtered out of the parsed depends table?
Or do you want the sdsm.json output to look like the output from calling export:matrix on a .dv8-sdsm file? I am assuming the former, following the diagram on https://github.com/sailuh/kaiaulu/issues/165. But I just wanted to make sure.

In particular, the sdsm.json output straight from the Depends command looks like this:
```
  "schemaVersion": "1.0",
  "name": "depends-output-sdsm",
  "variables": [...],
   "cells": [
    {
      "src": 1127,
      "dest": 60,
      "values": {
        "Call": 13,
        "Import": 1,
        "Use": 1,
        "Parameter": 1,
        "Contain": 2
      }
    },
    {
      "src": 702,
      "dest": 188,
      "values": {
        "Call": 27
      }
    },
    ...
```

Whereas the sdsm.json we get from calling core:export-matrix on a .dv8-sdsm looks like:
```
  "@schemaVersion" : "1.0",
  "name" : "depends-exported-matrix",
  "variables" : [...],
    "cells" : [ {
    "src" : 0,
    "dest" : 1,
    "values" : {
      "Use" : 7.0,
      "Call" : 1.0,
      "Create" : 1.0,
      "Contain" : 1.0
    }
  }, {
    "src" : 0,
    "dest" : 23,
    "values" : {
      "Use" : 3.0,
      "Import" : 1.0
    } ...
```

So the difference is the latter uses "@schemaVersion" instead of "schemaVersion", does not add "-sdsm" to the name of the project, and is sorted by src & dest.



> > 2. Should I let the user type the name of the project? I added this as a parameter because there is a "name" field in the sdsm.json with the name of the project, and I'm not sure how to get this otherwise.
> 
> Hmm..
> 
> On this issue:
> 
> [#168 (comment)](https://github.com/sailuh/kaiaulu/issues/168#issuecomment-1475972460)
> 
> ```
> tool:
>   # Depends allow to parse file-file static dependencies.
>    dv8:
>       # the project folder path files in function calls will be used. 
>       # The folder name will be used in the file names.
>       folder_path: ../../rawdata/dv8_thrift
>       # the architectural flaws thresholds that should be used
>       architectural_flaws:
>          clique_depends: call,use
>          crossing_cochange: 2
>          .
>          .
>          .
>          change_cover: ?            
> ```
> 
> We assume the user must specify a folder_path where all the data gets stored. Let's take that folder path containing the name of the project.
> 
> That being said, I think this assumption should be encoded on the R Notebook. I.e. the code to parse and extract the last folder in the path lives on the R Notebook. The path of the output file can then be constructed also on the notebook. Regarding the functions, I guess we can keep them simple and close to the commands from DV8. You expect an input path and output path variables where applicable. The R Notebook does the string concatenation for consistency when passing the parameters.
> 
> Does that make sense? That keeps the function more flexible if the user wants to use it for something else, and the R Notebook serves as the facilitator relying on the project configuration file.

I will add a dv8 config file to make this clearer and make the dv8 notebook more flexible. Thank you.


> 
> > 3. On the note of the "name" field, I tried to use auto_unbox to get it to be just a string value, but the json still generates the name values in a one-element list (see attached json called depends_to_sdsmj_result.json). Any suggestions how to fix this?
> 
> I am not sure what the auto_unbox function is. I believe you want to represent your json data in R as a list, since I'd expect jsonlite write function to map it 1 to 1 an R list to a json. If this doesn't answer the question can you give me an example of the issue or elaborate?

On this note, right now my dependencies_to_sdsmj() function is returning the sdsm.json as an R object. Would you rather I not return anything and just save the sdsm.json to an output folder?  

Also, I am using  in regards to auto_unbox, I use it on line 132 of my  dependencies_to_sdsmj() file in my commit https://github.com/sailuh/kaiaulu/pull/171/commits/8f82ab945010885dbaa4f79db9addc1aab6f518d#diff-f7557267e8cb3399084da6146186558b93329baa8fa5a87496e78935ac5e51eaR132

It looks like this:
```r
 return(jsonlite::fromJSON(jsonlite::toJSON(sdsm_json, auto_unbox = TRUE)))
```

The auto_unbox helped me change single-item lists in my json into just string values, except for the names field, which is still a single-item list. I will work on trying to fix this.


> 
> > 4. Do my parameters need to be in the exact order used by dv8?
> 
> This would be a question for Yi. I do not know how the source code of DV8 assumes the input, but I'd recommend staying consistent to the example files so any change in the future where DV8 cares about this, wouldn't break your function.

Okay, I will just stick to how the parameters are listed in the DV8 sdsm.json


> > 5. My biggest concern is the variables field in the sdsm.json. In particular, I thought this variable field would be a list composed of all the unique src and dest filenames. However, when I made a list of all the unique src and dest filenames from my parsed depends table, it was of smaller length than the variables array in the sdsm.json, by around 20-30 in size (1390 vs 1416). I'm still not sure what the difference is here.
> 
> Read both yours and the mismatching file variables as a character vector in R. Then use the function setdiff to identify the 20-30 that are unique.

I have done as you suggested. I compared the sdsm.json I get from my current dependencies_to_sdsmj() function and the sdsm.json from Depends directly. I had to replace the filepaths in the Depends sdsm.json so that it would match the same format as mine, where only the relevant part of the filepath is included, not the entire filepath.

```r
depends_json <- jsonlite::read_json("../../dv8-project/helix.json")
my_json <- jsonlite::read_json("../../dv8-project/checkingResultApril1.json")
```

```r
# Get depends sdsm.json unique variable names

file_names <- unlist(depends_json[["variables"]])
# /Users/user/git_repos/APR/xml/apr_xml_xmllite.c => "xml/apr_xml_xmllite.c"
folder_path = "/Users/xylent1/github/ics496/helix/"
depends_file_names <- stri_replace_first(file_names,replacement="",regex=folder_path)
depends_file_names
```

```r
# Get my sdsm.json unique variable names
my_file_names <- unique(my_json$variables)
```

```r
print(length(depends_file_names))
print(length(my_file_names))
setdiff(depends_file_names, my_file_names)
```

This is the output from doing setdiff:
```
[1] 1416
[1] 1390
 [1] "helix-common/src/main/java/org/apache/helix/ZNRecordDelta.java"                              
 [2] "helix-core/src/main/java/org/apache/helix/controller/package-info.java"                      
 [3] "helix-core/src/main/java/org/apache/helix/controller/pipeline/package-info.java"             
 [4] "helix-core/src/main/java/org/apache/helix/controller/stages/package-info.java"               
 [5] "helix-core/src/main/java/org/apache/helix/examples/package-info.java"                        
 [6] "helix-core/src/main/java/org/apache/helix/manager/zk/package-info.java"                      
 [7] "helix-core/src/main/java/org/apache/helix/messaging/handling/package-info.java"              
 [8] "helix-core/src/main/java/org/apache/helix/messaging/package-info.java"                       
 [9] "helix-core/src/main/java/org/apache/helix/model/package-info.java"                           
[10] "helix-core/src/main/java/org/apache/helix/monitoring/mbeans/package-info.java"               
[11] "helix-core/src/main/java/org/apache/helix/monitoring/package-info.java"                      
[12] "helix-core/src/main/java/org/apache/helix/package-info.java"                                 
[13] "helix-core/src/main/java/org/apache/helix/participant/package-info.java"                     
[14] "helix-core/src/main/java/org/apache/helix/participant/statemachine/package-info.java"        
[15] "helix-core/src/main/java/org/apache/helix/spectator/package-info.java"                       
[16] "helix-core/src/main/java/org/apache/helix/store/PropertyChangeListener.java"                 
[17] "helix-core/src/main/java/org/apache/helix/store/package-info.java"                           
[18] "helix-core/src/main/java/org/apache/helix/store/zk/PropertyItem.java"                        
[19] "helix-core/src/main/java/org/apache/helix/store/zk/ZkListener.java"                          
[20] "helix-core/src/main/java/org/apache/helix/store/zk/package-info.java"                        
[21] "helix-core/src/main/java/org/apache/helix/tools/commandtools/JmxDumper.java"                 
[22] "helix-core/src/main/java/org/apache/helix/tools/package-info.java"                           
[23] "helix-core/src/main/java/org/apache/helix/util/package-info.java"                            
[24] "helix-core/src/test/java/org/apache/helix/TestGetProperty.java"                              
[25] "recipes/rsync-replicated-file-system/src/main/java/org/apache/helix/filestore/ChangeLog.java"
[26] "recipes/rsync-replicated-file-system/src/main/java/org/apache/helix/filestore/Test.java"     
[1] 1416
[1] 1390
```
Any ideas on handling this?

> > EDIT: I have also made the changes you requested previously aside from vectorizing parse_dv8_clusters. I would like to return to that after finishing my milestone 3.4 functions.
> 
> I will take a look at your new commit. Thanks!
> 
> Keep in mind there are a number of changes that I am also going by guesswork here until I see the first full version of them. So we may need to revisit some assumptions I am proposing here after I see the first version, but I expect them to be mostly minimal like nomenclature.
> 
> Please point out if I missed anything.

Did you see my dependencies_to_sdsmj() and the dv8 notebook I had also added in a commit? Those are both works in progress (and I will redo the dv8 filepaths so they are using a dv8 config file instead of hardcoded paths), but just double-checking.

leilani-reich:
## Gitlog_to_hdsmj() concerns

On a separate note, I've also started looking at the gitlog_to_hdsmj() function. First, I inspected the hdsm json I got from running the following dv8 commands on a clone of the Apache Helix repo: gitlog --numstat  --date=iso  -->  scm:history:gittxt:convert-matrix --> core:export-matrix.

For this hdsm.json, I noticed that it is using the Cochange parameter. Do you know how we are defining Cochange? Is this the number of times the src and dest file were changed together at the same time for the same commit?

The format of the hdsm.json:
```
{
  "@schemaVersion" : "1.0",
  "name" : "gitlog-export-matrix",
  "variables" : [...],
"cells" : [ {
    "src" : 0,
    "dest" : 1,
    "values" : {
      "Cochange" : 3.0
    }
  }, {
    "src" : 0,
    "dest" : 2,
    "values" : {
      "Cochange" : 1.0
    }
...
```

carlosparadis:
Reply to (https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493258808) but we should probably break any message of this size in smaller independent messages.

> Hi Carlos, I have their emails. Actually, I was looking at the Depends repo and they show how the output json should look https://github.com/multilang-depends/depends/blob/master/doc/output_format.md. I have only seen "schemaVersion 1.0" used, but I will email about this to check.

0. This is very important for us as DV8 complies to it. Could you add it to our Wiki, a copy and paste of this, and reference the source in case it disappears in the future?  Just make a new section on this page: https://github.com/sailuh/kaiaulu/wiki/Depends

> So the difference is the latter uses "@schemaVersion" instead of "schemaVersion", does not add "-sdsm" to the name of the project, and is sorted by src & dest.

1. Let's go with the Depends implementation. However, include a sorting parameter that can take True or False and the functionality of sorting. Please also send Yi a second e-mail asking about this inconsistency, especially the sorting. DV8 can read both, correct? We want to make sure there are no adverse effects, especially involving the sorting, since our understanding of DV8 actual codebase is minimal.  

> This is the output from doing setdiff:


I need you to try and make the issues a bit more self-contained in your messages (I am finding myself scrolling back and forth and not being able to find what you are referring to). For example, your setdiff question and how to solve it, it took me at least 5 minutes going back and forth on the issue to remember that the problem was on `dependencies_to_sdsmj()`. It seems we are reaching the limit of GitHub interface itself (The discussion message board is a bit ahead of here but this has to leave on the issue tracker). 

I guess let's try breaking your next replies into separate messages if it needs to be for clarity. E.g. if they are all about the same function, keep them in one message and note them at the start of your message, but if they are different, break into 2 or more messages, but make sure you have the function being re-stated just in case at the start of them. If they are part of the same function but more elaborate, break them into separate messages too anyway for sanity sake. And I will match 1 separate message to every separate message for yours too hyperlinking what message it refers to. GitHub needs to urgently add the Discord "Reply to" button, since I am in essence asking us to start doing that manually. 

But to answer your question, are the files listed there being loaded in R? e.g. I imagine you are doing parse_dependencies() and subsequently using your function. Are the paths shown in R, and failing to be written in your function? I am not clear where in the pipeline the problem appears. 

> Did you see my dependencies_to_sdsmj() and the dv8 notebook I had also added in a commit? Those are both works in progress (and I will redo the dv8 filepaths so they are using a dv8 config file instead of hardcoded paths), but just double-checking.

No, I likely missed. I will take a look again. Feel free to remind again if unsure.

carlosparadis:
Reply to: https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493275804

This is the ideal size of a message for me at this point. Please feel free to flood the GitHub PR with messages like this in succession. It makes it much easier for me to read without all the scrolling, and I will do the same for you too. 

Also, when referring to the commands you use, use the function names we agreed upon. I have no idea what the commands are doing by DV8 nomenclature without having to look up that DV8 diagram I made, which is why I renamed them in our R functions. 

> gitlog --numstat --date=iso --> scm:history:gittxt:convert-matrix --> core:export-matrix.

I am assuming this means: dv8_gitlog_to_gitnumstat() -> dv8_gitnumstat_to_hdsmb() -> dv8_hdsmb_to_hdsmj()

> For this hdsm.json, I noticed that it is using the Cochange parameter. Do you know how we are defining Cochange? Is this the number of times the src and dest file were changed together at the same time for the same commit?

Yes. I was hopeful you would not need to do that yourself and it was done at a subsequent step of DV8. So you have to compute that metric. Try to do this:

1. Use parse_gitlog() 
2. Apply the transform gitlog_to_bipartite_network() function on mode="commit-file": http://itm0.shidler.hawaii.edu/kaiaulu/reference/transform_gitlog_to_bipartite_network.html
3. Use the bipartite_graph_projection(): ehttp://itm0.shidler.hawaii.edu/kaiaulu/reference/bipartite_graph_projection.html 

For mode, test both values of TRUE and FALSE, and go with the one that will give you a table of file_paths. It should have a weight column. That column value is the co-change metric. If this looks like sorcery to you, remember the Cheat Sheet you worked on, and which you documented this function. Look at the example graph where 1 of the 2 colored nodes are deleted, and the nodes it connected to are then connected. 

The co-change metric can be interpreted as a graph projection of file-commits. I suggest you sanity check the metric value matches to the one of DV8 too for a few pair of files. This will be a good way to also validate Kaiaulu's implementation. Use these functions within your function here, so it is self-contained and mimics the behavior of the dv8 command too.

I am doing some guesswork here that this should suffice. If something is not clear, please follow up right away on this co-change step since this is code you have not seen before and enters more the data analysis part.

leilani-reich:
Reply to https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493467231

# Updates on concerns surrounding dependencies_to_sdsmj()

> > Hi Carlos, I have their emails. Actually, I was looking at the Depends repo and they show how the output json should look https://github.com/multilang-depends/depends/blob/master/doc/output_format.md. I have only seen "schemaVersion 1.0" used, but I will email about this to check.
> 
> 0. This is very important for us as DV8 complies to it. Could you add it to our Wiki, a copy and paste of this, and reference the source in case it disappears in the future?  Just make a new section on this page: https://github.com/sailuh/kaiaulu/wiki/Depends

I added it to the wiki.

> > So the difference is the latter uses "@schemaVersion" instead of "schemaVersion", does not add "-sdsm" to the name of the project, and is sorted by src & dest.
> 
> 1. Let's go with the Depends implementation. However, include a sorting parameter that can take True or False and the functionality of sorting. Please also send Yi a second e-mail asking about this inconsistency, especially the sorting. DV8 can read both, correct? We want to make sure there are no adverse effects, especially involving the sorting, since our understanding of DV8 actual codebase is minimal.

I emailed Yi about the issue and cc-ed you, Rick, and Yuanfang.

leilani-reich:
Reply to: https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493467231

## dependencies_to_sdsmj(), unique filenames i get from the src & dest fields not completely matching the Depends sdsm.json output

> > This is the output from doing setdiff:
> 
> I need you to try and make the issues a bit more self-contained in your messages (I am finding myself scrolling back and forth and not being able to find what you are referring to). For example, your setdiff question and how to solve it, it took me at least 5 minutes going back and forth on the issue to remember that the problem was on `dependencies_to_sdsmj()`. It seems we are reaching the limit of GitHub interface itself (The discussion message board is a bit ahead of here but this has to leave on the issue tracker).
> 
> I guess let's try breaking your next replies into separate messages if it needs to be for clarity. E.g. if they are all about the same function, keep them in one message and note them at the start of your message, but if they are different, break into 2 or more messages, but make sure you have the function being re-stated just in case at the start of them. If they are part of the same function but more elaborate, break them into separate messages too anyway for sanity sake. And I will match 1 separate message to every separate message for yours too hyperlinking what message it refers to. GitHub needs to urgently add the Discord "Reply to" button, since I am in essence asking us to start doing that manually.
> 
> But to answer your question, are the files listed there being loaded in R? e.g. I imagine you are doing parse_dependencies() and subsequently using your function. Are the paths shown in R, and failing to be written in your function? I am not clear where in the pipeline the problem appears.

The files that I am missing in my function seem to disappear from the table produced by parse_dependencies(). But they are in the sdsm.json generated by the Depends command.

> 
> > Did you see my dependencies_to_sdsmj() and the dv8 notebook I had also added in a commit? Those are both works in progress (and I will redo the dv8 filepaths so they are using a dv8 config file instead of hardcoded paths), but just double-checking.
> 
> No, I likely missed. I will take a look again. Feel free to remind again if unsure.

leilani-reich:
Reply to: https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493471686

## gitlog_to_hdsmj()

> This is the ideal size of a message for me at this point. Please feel free to flood the GitHub PR with messages like this in succession. It makes it much easier for me to read without all the scrolling, and I will do the same for you too.
> 
> Also, when referring to the commands you use, use the function names we agreed upon. I have no idea what the commands are doing by DV8 nomenclature without having to look up that DV8 diagram I made, which is why I renamed them in our R functions.
> 
> > gitlog --numstat --date=iso --> scm:history:gittxt:convert-matrix --> core:export-matrix.
> 
> I am assuming this means: dv8_gitlog_to_gitnumstat() -> dv8_gitnumstat_to_hdsmb() -> dv8_hdsmb_to_hdsmj()

Yes, I will use the nomenclature from now on and take the message size into account.

> > For this hdsm.json, I noticed that it is using the Cochange parameter. Do you know how we are defining Cochange? Is this the number of times the src and dest file were changed together at the same time for the same commit?
> 
> Yes. I was hopeful you would not need to do that yourself and it was done at a subsequent step of DV8. So you have to compute that metric. Try to do this:
> 
> 1. Use parse_gitlog()
> 2. Apply the transform gitlog_to_bipartite_network() function on mode="commit-file": http://itm0.shidler.hawaii.edu/kaiaulu/reference/transform_gitlog_to_bipartite_network.html
> 3. Use the bipartite_graph_projection(): ehttp://itm0.shidler.hawaii.edu/kaiaulu/reference/bipartite_graph_projection.html
> 
> For mode, test both values of TRUE and FALSE, and go with the one that will give you a table of file_paths. It should have a weight column. That column value is the co-change metric. If this looks like sorcery to you, remember the Cheat Sheet you worked on, and which you documented this function. Look at the example graph where 1 of the 2 colored nodes are deleted, and the nodes it connected to are then connected.
> 
> The co-change metric can be interpreted as a graph projection of file-commits. I suggest you sanity check the metric value matches to the one of DV8 too for a few pair of files. This will be a good way to also validate Kaiaulu's implementation. Use these functions within your function here, so it is self-contained and mimics the behavior of the dv8 command too.
> 
> I am doing some guesswork here that this should suffice. If something is not clear, please follow up right away on this co-change step since this is code you have not seen before and enters more the data analysis part.

I will look into getting the cochange, thank you.

carlosparadis:
Reply to: https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493750703

1. Ok, I understand now. The problem is again on parse_dependencies() then. Could you open another issue, and make this problem self-contained there (so I don't need to refer this PR for details), and assign it to me, so I can look into it later? I believe you will need to send me the folder of the code you are using too. Try attaching it to the issue to see if it goes through. Otherwise, please send me by e-mail and mention the new issue number. 

2. For now, it seems you are able to generate the missing file paths. Would it be doable for you to manually add these missing filepaths to your table, and then call your function to see if it matches to the correct file? (Just so you can test your function on the assumption parse_dependencies() is working). 

3. Also, have you checked DV8 can ingest your currently generated file?  I think it would be good if you send me your currently created DSM file (even with missing files), so I can try to load it on DV8 GUI too. You can attach said file to this issue in your reply. 

In the meantime, if you are confident the problem lies in parse_dependencies(), please move ahead with the other tasks. Thanks!

leilani-reich:
Reply to https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493889187



> Reply to: [#171 (comment)](https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493750703)
> 
> 1. Ok, I understand now. The problem is again on parse_dependencies() then. Could you open another issue, and make this problem self-contained there (so I don't need to refer this PR for details), and assign it to me, so I can look into it later? I believe you will need to send me the folder of the code you are using too. Try attaching it to the issue to see if it goes through. Otherwise, please send me by e-mail and mention the new issue number.

I have made a new issue https://github.com/sailuh/kaiaulu/issues/178 and attached a zip with relevant files.

> 2. For now, it seems you are able to generate the missing file paths. Would it be doable for you to manually add these missing filepaths to your table, and then call your function to see if it matches to the correct file? (Just so you can test your function on the assumption parse_dependencies() is working).

Yes, I can test that out.

> 3. Also, have you checked DV8 can ingest your currently generated file?  I think it would be good if you send me your currently created DSM file (even with missing files), so I can try to load it on DV8 GUI too. You can attach said file to this issue in your reply.

The DV8 command seems to have worked for my json. I have attached a zip with my sdsm.json and the outputted dsm.
[dependencies_to_sdsmj_output.zip](https://github.com/sailuh/kaiaulu/files/11137363/dependencies_to_sdsmj_output.zip)


> 
> In the meantime, if you are confident the problem lies in parse_dependencies(), please move ahead with the other tasks. Thanks!
Got it. I will work on the gitlog_to_hdsmj() and revisit the bugzilla functions.

leilani-reich:
Reply to https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1493467231

## Info on sdsm.json format from YI

> > So the difference is the latter uses "@schemaVersion" instead of "schemaVersion", does not add "-sdsm" to the name of the project, and is sorted by src & dest.
> 
> 1. Let's go with the Depends implementation. However, include a sorting parameter that can take True or False and the functionality of sorting. Please also send Yi a second e-mail asking about this inconsistency, especially the sorting. DV8 can read both, correct? We want to make sure there are no adverse effects, especially involving the sorting, since our understanding of DV8 actual codebase is minimal.

### Hi Carlos, did you see my email to Yi? I will add it here for reference:

1. I am working on a function to recreate the sdsm json from a table of dependencies, and I was wondering what the value for the json field “schemaVersion" should be.
```
schemaVersion is a placeholder for future extension, for now you could use same for both sdsm and hdsm.
```
2. Also, I noticed for the sdsm.json, each “values” field contains an object with different fields for dependency types (ex. Call, Import, Create, …). Do I need to follow the same order the dependency types are written for recreating the sdsm.json?
```
No it's not necessary. Actually the ordering rules are different for hash print depending on the programming languages. But it's not that harmful.
```
3. I am also working on a function to recreate the hdsm json from a gitlog table. In regards to the “Cochange” field, I wanted to clarify what this means. Is this the number of times that the particular src and dest pair get changed together, which we identify by looking at how many times they are committed together in the gitlog table?
```
Your understanding is correct.
```
4. In addition, should the values in the “cells" field in the dsm.jsons I recreate be sorted by src & dest pair? In the sdsm.json I get straight from Depends, it does not seem to sort the cells this way. However, if I export either an sdsm.dv8-dsm or hdsm.dv8-dsm to a dv8.json, the cells are sorted by src & dest pair.
```
It's also unnecessary to sort the "cells" field, that doesn't impact anything.
```
###  Summary of email
So to summarize, I can use schemaVersion 1.0 for both the hdsm and sdsm jsons, and the order of the parameters for the values field in the sdsm.json does not matter, nor does the sorting of the objects in the "cells" field (don't need to sort by src & dest). Do you still want me to add a sorting parameter for my dependencies_to_sdsmj() function to sort the "cells" field values by src & dest?

carlosparadis:
Reply to https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1495588594

Yes, I saw the exchange. If it is not too much work, include it as an optional parameter for manual inspection/troubleshooting sake. Otherwise, you can skip.

leilani-reich:
Hi Carlos,

I've added the dependencies_to_sdsmj() and gitlog_to_hdsmj(). 
I also added an is_sorted parameter for both functions so if a user wanted to sort by src and dest they could.

Some notes/concerns on the functions:

## dependencies_to_sdsmj()
I compared the output of dependencies_to_sdsmj() (after adding back the missing variables that we need to address in https://github.com/sailuh/kaiaulu/issues/178) to the output of the sdsm.json we get from calling the Depends command -> dv8_dsmj_to_dsmb() -> sdsm.dsm -> dv8_dsmb_to_dsmj(), and it looks the same aside from the order of the parameters in the values field. However, that order does not seem consistent in the file produced by dv8, so I didn't try making a particular order of these parameters. So I'm confident in my implementation of this function.

### Output
I will attach the output I got from my dependencies_to_sdsmj versus DV8's output for the Depends.json. For Depends, I used the Apache Helix project.

[depends-output-april-9.zip](https://github.com/sailuh/kaiaulu/files/11187789/depends-output-april-9.zip)


## gitlog_to_hdsmj()
For this function, I am running into an issue where again the filenames from my function do not match the dv8 output (running dv8_gitlog_to_numstat() -> dv8_gitnumstat_to_hdsmb() -> dv8_dsmb_to_dsmj()). By filenames I am referring to the "variables" field in the hdsm.json. I took the filenames from the "from" and "to" columns in the table we get from running the following: parse_gitlog() -> transform_gitlog_to_bipartite_network() -> bipartite_graph_projection(gitlog_graph, mode = FALSE, is_intermediate_projection = FALSE) -> cochange_table[[2]]. I show this on the dv8.R in my latest commit on this pull request.

In particular, sometimes I am getting more filenames than there are in the output DV8 produces. Other times I am getting less filenames than DV8 produces. I am confused as to why this is happening. Can you advise me on this?

### Output
I will attach the output I got from my gitlog_to_hdsmj versus DV8's output for the gitlog.json. For this, I tested a small project called [QtNotepad](https://github.com/rattle99/QtNotepad) and a project I worked on called [Bridging The Gap](https://github.com/bridging-the-gap/bridging-the-gap). I tried these projects because they were relatively small and gitlog_to_hdsmj took a long time for the regular helix repo. I also tried just using a smaller version of helix using git checkout, but I'm not sure if I did that right, so to be safe I tried these small projects.

[gitlog-output-april-9.zip](https://github.com/sailuh/kaiaulu/files/11187792/gitlog-output-april-9.zip)

## Extra notes
I removed the dv8 notebook I talked about because I think that should be on an issue of its own. Plus, it's using hard-coded paths. I am also still working on trying to vectorize parse_dv8_clusters().

Thanks,
Leilani

carlosparadis:
Reply to https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1501397379 

> dependencies_to_sdsmj()

This is fine. I tested on dv8-gui loading the file and it works. I am experiencing a strange error, when I click a few things, but dv8 file also generates it. So I will check on Yi on that. But otherwise, I believe you succeeded in replicating the conversion, pending the bug I have to check.

> gitlog_to_hdsmj()

> In particular, sometimes I am getting more filenames than there are in the output DV8 produces. Other times I am getting less filenames than DV8 produces. I am confused as to why this is happening. Can you advise me on this?

We will need to narrow down again, like last time, where more filenames are added in the chain. For starters, what do you mean sometimes it is more, and sometimes it is less? Are you saying running over the same data every time gives you a different number of paths, --or-- that in different datasets it can gives more or less? Do you have an example of each case?

I suspect the issue here may lie on how parse_gitlog() and DV8 include branches.

Also, just to be clear, you are --not-- using the filter functions when generating these files right? DV8 should not be filtering anything, so we should not use them.

https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L42-L53

These flags are due getting moved to the project configuration file at some point. Could you confirm to me if, after this code line:

https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/parser.R#L56-L61

The file names match between DV8 and Perceval? I.e. is Perceval data correct, and some subsequent step causes the mismatch? Otherwise, it may simply be the way we are exporting the git log via --numstat for DV8, and via Perceval that may be causing the differences. 

For sanity sake, please move this into a separate issue, much like the Depends one. You can quote this response there, and we can follow there. Let's keep this PR strictly about any concerns of your function implementations to minimize confusion. 

Note I am posting this before reviewing your last commits code.

carlosparadis:
On the subject of the Pull Request, has your group decided on the rstudioapi:: call yet?

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+  cluster_parsed <- data.table::setnames(cluster_parsed, c("V1", "V2", "V3"), c("filename", "module", "layer"))
+  # Return parsed cluster data table
+  return(cluster_parsed)
+}
+
+#' Convert dependencies table to a sdsm.json
+#'
+#' Converts table of dependencies from Depends into an *-sdsm.json binary.
+#'
+#' @param depends_table parsed dependencies table
```
carlosparadis, line 86: Use \code{\link{}} to parse_dependencies, and add to @seealso

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+#' @param project_name project name for the sdsm.json
```
carlosparadis, line 87: Hmm... I may be contradicting earlier feedback here, but we may want to call this variable sdmj_path for consistency here. 

The Notebook, however, should handle the logic of obtaining the project name out of the project configuration file, no? Please let me know if this conflicts to something we discussed prior.

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+    depends_src <- depends_table[["src"]][[i]]
+    depends_dest <- depends_table[["dest"]][[i]]
+
+    # Look at first value in parse_depends table, get matching index in vars array
+    src_index <- variables_indices[depends_src] - 1
+    dest_index <- variables_indices[depends_dest] - 1
+
+    values = list()
+    parameters = list("Call", "Import", "Use", "Parameter", "Contain", "Create", "Extend",
+                      "Return","Implement", "Cast", "Throw", "Annotation")
```
carlosparadis, line 124: Does your function assume all types of parameters are always present? Or would it work if a user, for example, subset the output of parse_dependencies() only to contain a subset of them and then provides to your function?

It is very important the latter works, as this would enable full customization of the analysis for dv8 should the user wish to remove types of dependencies in their analysis, and the reason why this is included on config files:

https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/conf/thrift.yml#L101-L115

To be clear, the subsetting is done at Notebook level, instead of function level. I just want to make sure your function does not presume all dependencies are always present (which is also impossible to guarantee depending on the project being parsed).

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+
+  # Rename result columns
+  cluster_parsed <- data.table::setnames(cluster_parsed, c("V1", "V2", "V3"), c("filename", "module", "layer"))
+  # Return parsed cluster data table
+  return(cluster_parsed)
+}
+
+#' Convert dependencies table to a sdsm.json
+#'
+#' Converts table of dependencies from Depends into an *-sdsm.json binary.
```
carlosparadis, line 84: Could you expand the description of this function to include more information on the expected architecture of the json that you found and added to the wiki? I think it would be very helpful for if changes are required in the future that the docs of this function on the body errs on being verbose.

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+  cells_df <- data.table::data.table(jsonlite::fromJSON(jsonlite::toJSON(cells, auto_unbox = TRUE)))
+
+  # Sort by the src_index and dest_index if is_sorted is true
+  if (is_sorted==TRUE){
+    data.table::setorder(cells_df, cols = "src", "dest")
+  }
+
+  sdsm_json <- list(schemaVersion="1.0", name=paste0(project_name, "-sdsm"), variables=variables, cells=cells_df)
+
+  json_df <- jsonlite::fromJSON(jsonlite::toJSON(sdsm_json, auto_unbox = TRUE))
```
carlosparadis, line 148: Could you check if read_json could be used here? I believe toJSON makes some implicit transformations. Likewise for any other use of this function.

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+  # Unbox each of the cell values (should be values: object, not values: [object])
+  json_df$cells$values <- lapply(json_df$cells$values, jsonlite::unbox)
+
+  # Save the json to a file
+  jsonlite::write_json(json_df, paste0("../rawdata/dv8/", project_name, ".json"), auto_unbox=TRUE)
+}
+
+#' Convert gitlog table to a hdsm.json
+#'
+#' Converts a gitlog table into an *-hdsm.json binary.
```
carlosparadis, line 159: Same request here on expanding on the body docs and err on verbose end given the large number of assumptions and schema of the file. The other functions besides the two that write jsons are ok to be short.

carlosparadis (changes requested):
These look good to me. Some minor requests on code and more docs on the json writers since they encode a lot of your understanding of how these files are parsed to DV8. I know they are discussed on our issues and Pull Requests in length, but I would like to distill that information into the body of these two functions so they are easily found in the future.

As noted in my other comment, I guess the major concern is the pinpoint of the issue that is affecting your hdsmj writer and I figuring a fix for the filepaths of the sdsmj writer so we can make sure everything works as intended. 

Please branch a new issue for hdsmj writer problem so we iterate on that once you can pinpoint more precisely where the problem lies.

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+#'
+#' @param gitlog_table parsed gitlog table
+#' @param project_name project name for the hdsm.json
+#' @param is_sorted whether the json is sorted by src and dest
+#' @export
+#' @family dv8
+gitlog_to_hdsmj <- function(gitlog_table, project_name, is_sorted=FALSE){
+  # Call preliminary functions to get graph and cochange for the files
+  gitlog_graph <- transform_gitlog_to_bipartite_network(gitlog_table, mode ="commit-file")
+  cochange_table <- bipartite_graph_projection(gitlog_graph, mode = FALSE, is_intermediate_projection = FALSE)
```
carlosparadis, line 169: Were you able to check the values of co-change here match to DV8?

carlosparadis: Also curious if this is the source of the slow down on the function brought up. Could you try running line by line to see where is the performance bottleneck? 

Same ask for the dependencies_sdsmj(). Where is the bottleneck?

carlosparadis:
Reply to: https://github.com/sailuh/kaiaulu/pull/171#issuecomment-1503153951

>  QtNotepad, my gitlog_to_hdsmj() outputs a json with 8 variables whereas the json output by dv8 gives 10 variables.

Could you send me this and the other files so I can take a look? QtNotepad is a great starting point since it is so small.

R/dv8.R
carlosparadis: > which variable would be sdmj_path? depends_table? I thought we would take as input the table object from parse_dependencies and not a filepath?

No, `project_name`.

R/dv8.R
leilani-reich: Hi Carlos, my function does not assume all parameters are present.
>Or would it work if a user, for example, subset the output of parse_dependencies() only to contain a subset of them and then provides to your function?
Yeah that should work fine. I think for most cells in my json they don't have all dependencies anyway.

Also, referring to my dependencies_to_sdsmj()
```r
    # Get all the parameter values for a specific cell
    for (j in 1: length(parameters)) {
      current_param <- depends_table[i][[parameters[[j]]]]
      if (current_param > 0){
        values[[parameters[[j]]]] <- current_param
      }
    }
```

In the if statement, I check that the parameter I am currently looking at for a particular cell exists, and only then do I add it to the values for a cell.

carlosparadis: Then it works as intended for me! :)

R/dv8.R
leilani-reich: Hi Carlos, the values of co-change did not exactly match, likely since my output file has a different number of variables than the dv8 output file https://github.com/sailuh/kaiaulu/issues/184.

>Also curious if this is the source of the slow down on the function brought up. Could you try running line by line to see where is the performance bottleneck?

I don't believe transform_gitlog_to_bipartite_network() nor bipartite_graph_projection() were the bottleneck, running these lines individually.

Well, actually the table generated by bipartite_graph_projection() may be the problem. I got a huge table when I ran the original Apache Helix repository on my gitlog_to_hdsmj() function. I don't remember how large it was, but the json it generated crashed my RStudio for a while, leaving me with a white screen whenever I tried to open the app. Had to manually move/delete a file to fix that.

>Same ask for the dependencies_sdsmj(). Where is the bottleneck?

I don't think there's a bottleneck for this function. The only one I noticed being slow was gitlog_to_hdsmj().

R/dv8.R
```diff
@@ -0,0 +1,228 @@
...
+  # Get the files in the from (src) and to (dest) columns, concatenate them into one list, sort alphabetically
+  # Getting files in the "from" column
+  gitlog_files_src <- unique(cochange_table[["from"]])
+  # Getting files in the "to" column
+  gitlog_files_dest <- unique(cochange_table[["to"]])
+  # Combining src and dest columns
+  gitlog_files <- c(gitlog_files_src, gitlog_files_dest)
+
+  # Taking the unique combined columns
+  variables <- unique(gitlog_files)
```
carlosparadis, line 183: @leilani-reich This, paired with your assumption of line 172, 

```r
 # List of two tables. 2nd table had the filenames and cochange
  cochange_table <- cochange_table[[2]]
```

is where the second part of the "variables are missing" is introduced in #184. More specifically, variables should be pulled from `cochange_table[[1]]`. I'd also recommend you rename these variables here for clarity:

cochange_table[[1]] should be referred to as `nodes_table`, where as `cochange_table[[2]]` should be referred to as `edgelist_table`.

Here's what `nodes_table` looks like, after the fix of part 1 in #184 (not committed yet at the time of this comment):

```
                        name  type   color
 1:              .gititgnore FALSE #f4dbb5
 2:                README.md FALSE #f4dbb5
 3:  QtNotepad/QtNotepad.pro FALSE #f4dbb5
 4:       QtNotepad/main.cpp FALSE #f4dbb5
 5: QtNotepad/mainwindow.cpp FALSE #f4dbb5
 6:   QtNotepad/mainwindow.h FALSE #f4dbb5
 7:  QtNotepad/mainwindow.ui FALSE #f4dbb5
 8:  QtNotepad/resources.qrc FALSE #f4dbb5
 9:               .gitignore FALSE #f4dbb5
10:                  LICENSE FALSE #f4dbb5
11:               LICENSE.md FALSE #f4dbb5
```

As you can see, the all the files are listed here. 

Let me clarify what structure you are looking at for "cochange_table": When a graph (the circles with connections of your Social Smell cheatsheet) is represented in memory, you have a few options. One of them is representing them as a pair of tables: Nodes, which defines all information a node will have (e.g. the color of a node, it's size, etc), 

and an edgelist table. This edgelist table defines only the connections, which happens in **pairs**. In this code, you assumed that the edgelist contains all the nodes, but that assumption is not always true. 

Indeed, in this project, LICENSE.md is committed only once (during a file rename), and more importantly, it is committed without any other file. This leads to LICENSE.md having no connections to any other file in the entire of this git log, and thus will not appear on the edgelist table. **But it will appear on the node table**. 

This format of node and edgelist tables is also what DV8 is using: Note in the file you sent me (in this comment https://github.com/sailuh/kaiaulu/issues/184#issue-1662363636), LICENSE.md appears under "variables" (DV8's node list table), but not in the "cells" (DV8's edgelist table).

R/dv8.R
carlosparadis: Just as a reminder from our last meeting: We want to make sure your function here can account for parse_dependencies() having less columns (as every column represents a type in the configuration file) should the user decide to include less types:

https://github.com/sailuh/kaiaulu/blob/d73c603a2849b5749e60ea42c977c925a489e1b0/conf/apr.yml#L50-L63

You can see an example of the table here as a reminder: http://itm0.shidler.hawaii.edu/kaiaulu/articles/depends_showcase.html#example-network-visualization-with-igraph-1

Otherwise, there will be no way to leverage the user choice when passing it to DV8. Let me know if this doesn't make sense.

carlosparadis:
hi, is this graph and transform functions functional (aside from the issues we already know))? I wanted to do the merge and then we could refine from there those specifics in a new issue.

R/download.R
```diff
@@ -1,9 +1,3 @@
-# Kaiaulu - https://github.com/sailuh/kaiaulu
-#
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at https://mozilla.org/MPL/2.0/.
-
```
carlosparadis, line 6: Not sure why this was removed... but I guess I will just re-add after the merge.

carlosparadis (approved):
Functions worked on DV8 Notebook. Minor documentation issues. The R/download.R copyleft has to be re-added. Not sure why it was removed. 

Was hoping to merge before the addition of network/R and graph/R in progress functions but that is fine, since the others from 3.4 Milestone are still around. We can work on refining them in a follow-up issue. 

In retrospect, Milestone 3.4 functions, given their complexity and reliance on other Kaiaulu functions which bugs were identified, should have likely be a separate PR to avoid this one growing too long.

Issue #153, opened 2023-02-16
Bugzilla Crawler


leilani-reich:
### Proposal

We will implement a Bugzilla crawler that will be directly called from R. Bugzilla has a REST API, so we can parse the JSON that is returned by the API call in the R function.

The main contributions for this issue will be the Bugzilla crawler functions, unit tests for these functions, and a vignette demonstrating how to use the Bugzilla crawler.

### Architecture

NOTE: We should discuss the architecture of the Bugzilla crawler before starting any code.

As stated by Carlos:
The issues raised on issue #70 of PERCEIVE provide good motivation on why we should refine the requirements of it upfront: https://github.com/sailuh/perceive/issues/70. 

More specifically, we will want to break the Bugzila crawler into a 
- i) "downloader as-is" (which I believe the raw data is json)
- ii) parser of said json into tables. 

The downloader would be under R/download.R, where as the crawler under R/parser.R.

carlosparadis:
This may prove useful:

Possibly useful Docs:
- https://wiki.mozilla.org/Bugzilla:REST_API
- https://bz.apache.org/ooo/docs/en/html/integrating/api/Bugzilla/WebService/Server/REST.html

E.g. Bug 35: https://bz.apache.org/bugzilla/show_bug.cgi?id=35
Bug 35 via REST interface: https://bz.apache.org/bugzilla/rest.cgi/bug/35

Also, please let me know which R libraries you believe are needed for this task. Ideally, the ones already listed as dependencies should suffice:

https://github.com/sailuh/kaiaulu/blob/c968a6ded22610db52e9aaa261d258d5d7128976/DESCRIPTION#L34-L45 

R/download.R is a good start for examples using them for downloading data, and R/parse.R for parsing input from Perceval via system2 calls.

leilani-reich:
Hi, here's a rough function for the downloader. It get's the data using the REST API, and will start bug retrieval based on the date parameter. There is an option to download comments for the bugs if the user wants. The function will output a list from which users can obtain issue data and comment data in json format.

```r
#'  Download Bugzilla issues and comments using the Bugzilla REST API
#'
#' @param bugzilla_site link to specific bugzilla site
#' @param date the date to start bug retrieval
#' @param comments whether to download comments
#' @export
download_bugzilla_from_rest_api <- function(bugzilla_site, date, comments=FALSE){
  # Format link to retrieve data using Bugzilla REST API
[...]

PR #177, opened 2023-04-03
Kaiaulu's bugzilla Crawler Downloader

leilani-reich:
I have updated my downloader function according to feedback on https://github.com/sailuh/kaiaulu/issues/153#issuecomment-1482339183. 

The changes added are mainly in R/downloader.R.  I also updated NEWS.md, _pkgdown.yml, CONTRIBUTING.md, and the NAMESPACE.

I was trying to get rid of merge commits with a git squash, but the git squash doesn't show merge commits, so I did a git reset to make only a single commit for my bugzilla downloader. However, I think it shows a lot more unnecessary stuff now. Sorry about that. I can redo it if that's better.


carlosparadis:
Haha, indeed 31 files. Could you let me know which files you want me to review? R/download.R for certain, but what were the others? I can work with that.

leilani-reich:
Yes, it is just the R/downloader functions download_bugzilla_issues_from_rest_api() and download_bugzilla_comments_from_rest_api(). And the NEWS.md, _pkgdown.yml, CONTRIBUTING.md, and the NAMESPACE, and the man files for the two downloader functions.

Actually, I think I added the bugzilla_wrapper_showcase notebook by accident. I will double-check what I added and redo the commit if that's ok.

carlosparadis:
Yes, that's fine. Go ahead and make the needed changes and post another comment when ready to go :)

leilani-reich:
Hi Carlos, I removed the file that shouldn't have been in my commit. Thanks!

R/download.R
```diff
@@ -4,6 +4,75 @@
...
+#' @param start_timestamp when to start bug retrieval (ex. 2023-01-01T00:14:57Z)
+#' @param save_folder_path the full *folder* path where the bugzilla issues will be stored
+#' @seealso \code{\link{download_bugzilla_comments_from_rest_api}}
+#' @export
+download_bugzilla_issues_from_rest_api <- function(bugzilla_site, start_timestamp, save_folder_path){
+  # Format link to retrieve data using Bugzilla REST API
+  bugzilla_site <- paste(bugzilla_site, "/rest", sep="")
+
+  # Make sure folder path is correctly formatted
+  if (stringr::str_sub(save_folder_path,-1) != "/"){
```
carlosparadis, line 22: Kaiaulu rely on the stringr library. Could you replace this by stringi library? There should be a function equivalent to this one.

R/download.R
```diff
@@ -4,6 +4,75 @@
...
+                      httr::write_disk(file.path(paste0(save_folder_path, 0, ".json")), overwrite = TRUE))
+
+  bug_ids = c()
+  offset = 20
+  page = 0
+  # Get the bugs and save each page to a json file in the specified save_folder_path
+  while(httr::status_code(issues) == 200){
+    issues = httr::GET(paste(bugzilla_site, "/bug", "?creation_time=", start_timestamp, sep=""),
+                       limit=20, offset=offset,
+                       httr::write_disk(file.path(paste0(save_folder_path, page, ".json")), overwrite = TRUE))
```
carlosparadis, line 38: This is very good!

R/download.R
```diff
@@ -4,6 +4,75 @@
...
+  if (stringr::str_sub(save_folder_path,-1) != "/"){
+    save_folder_path <-paste0(save_folder_path, "/")
+  }
+
+  # Get the issues starting from the specified date
+  issues <- httr::GET(paste(bugzilla_site, "/bug", "?creation_time=", start_timestamp, sep=""),
+                      limit=20, offset=0,
+                      httr::write_disk(file.path(paste0(save_folder_path, 0, ".json")), overwrite = TRUE))
+
+  bug_ids = c()
```
carlosparadis, line 31: Please use <- for consistency

R/download.R
```diff
@@ -4,6 +4,75 @@
...
+  # Make sure folder path is correctly formatted
+  if (stringr::str_sub(save_folder_path,-1) != "/"){
+    save_folder_path <-paste0(save_folder_path, "/")
+  }
+
+  # Get the comments associated with the bug_ids and save each group of comments for a bug
+  # to a json file at the specified save_folder_path
+  for (i in 1:length(bug_ids)){
+    comments <- httr::GET(paste(bugzilla_site, "/bug/", bug_ids[i], "/comment", sep=""),
+                          httr::write_disk(file.path(paste0(save_folder_path, bug_ids[i], ".json")), overwrite = TRUE))
```
carlosparadis, line 72: Very good!

carlosparadis (changes requested):
Only a very minor change on the function call of `stringr` to `stringi`, otherwise it seems fine to me! 

Did you test the code and got it download some data? Could you add a screenshot to the PR of how your folder looks like after running both?

R/download.R
```diff
@@ -4,6 +4,75 @@
...
 
+#' Download Bugzilla issues using the Bugzilla REST API
+#'
+#' Downloads bugzilla issues and returns a list of bugzilla bug ids for use with
+#'  \code{\link{download_bugzilla_comments_from_rest_api}}
+#'
+#' @param bugzilla_site URL to specific bugzilla site
+#' @param start_timestamp when to start bug retrieval (ex. 2023-01-01T00:14:57Z)
+#' @param save_folder_path the full *folder* path where the bugzilla issues will be stored
+#' @seealso \code{\link{download_bugzilla_comments_from_rest_api}}
```
carlosparadis, line 15: Could you also include Malia's associated parser function to this one?

R/download.R
```diff
@@ -4,6 +4,75 @@
...
+}
+
+#' Download Bugzilla comments using the Bugzilla REST API
+#'
+#' Downloads comments associated with each bug id returned from \code{\link{download_bugzilla_issues_from_rest_api}}
+#'
+#' @param bugzilla_site URL to specific bugzilla site
+#' @param bug_ids the ids of the bugs to extract comments for from \code{\link{download_bugzilla_issues_from_rest_api}}
+#' @param save_folder_path the full *folder* path where the bugzilla comments will be stored
+#' @seealso \code{\link{download_bugzilla_issues_from_rest_api}}
```
carlosparadis, line 57: Could you also include Malia's associated parser function to this one?

carlosparadis:
Reply to https://github.com/sailuh/kaiaulu/pull/177#issuecomment-1502714303 

Ok! Looks good :) Just to be clear, the file names for the issues are the pages, which contain a group of topics, and for the comments, each file is a comment, and the file name is a comment id, correct?

Also, after you make the final changes, could you try testing the final version against just another bugzilla project to see if it works too? That way we are sure we didn't hardcode anything too specific for Red Hat.

R/download.R
carlosparadis: Also, you may want to see what Perceval does with their restful API endpoint. Do they ask a parameter from the user? Surely they had to also decide on this question!

carlosparadis:
Note to self so I don't forget: 

This is pending so we can accept the PR: https://github.com/sailuh/kaiaulu/pull/177#pullrequestreview-1378369733

carlosparadis:
Reply to https://github.com/sailuh/kaiaulu/pull/177#issuecomment-1502829995 

Got it. Could you include the answer to my questions as part of the body of the two functions too? I think this will be helpful for the users!

R/download.R
```diff
@@ -4,6 +4,74 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+#' Download Bugzilla issues using the Bugzilla REST API
+#'
+#' Downloads bugzilla issues and returns a list of bugzilla bug ids for use with
+#'  \code{\link{download_bugzilla_comments_from_rest_api}}
```
carlosparadis, line 10: Add how files are saved here: https://github.com/sailuh/kaiaulu/pull/177#issuecomment-1502829995

R/download.R
```diff
@@ -4,6 +4,74 @@
...
+    offset <- offset + 20
+    page <- page + 1
+  }
+
+  return(bug_ids)
+}
+
+#' Download Bugzilla comments using the Bugzilla REST API
+#'
+#' Downloads comments associated with each bug id returned from \code{\link{download_bugzilla_issues_from_rest_api}}
```
carlosparadis, line 51: Add how files are saved here: https://github.com/sailuh/kaiaulu/pull/177#issuecomment-1502829995

R/download.R
leilani-reich: Hi Carlos,
I was looking at https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/github.R#L329-L350

I didn't see an exact answer to using a different condition for the while loop, but to solve the problem where I end up writing an empty json file at the end of my download issues function, I was wondering if I could use write_json instead of htrr:write_disk? That way, I can separate the GET request from actually writing the json to a file. I tried it out and I can do it so that I don't generate an empty file at the end.


```r
  while(length(httr::content(issues)$bugs) > 0){
    issues <- httr::GET(paste(bugzilla_site, "/bug", "?creation_time=", start_timestamp,
                             "&limit=20", "&offset=", offset, sep=""))

    if (length(httr::content(issues)$bugs) == 0){
      print("Break so we don't write empty file to folder.")
      break
    }

    # Writing the file to a json, separate from the GET request
    jsonlite::write_json(httr::content(issues), paste0(save_folder_path, page, ".json"), pretty=TRUE, auto_unbox=TRUE)

    json_issues <- rawToChar(httr::content(issues, as = "raw"))
    current_bug_ids <- jsonlite::fromJSON(json_issues)[["bugs"]][["id"]]
    bug_ids <- c(bug_ids, current_bug_ids)
    offset <- offset + 20
    page <- page + 1
  }
```

leilani-reich: >Interesting. Did you check the API documentation to see if there isn't a command where you can pull the limit set by the administrator? Also could you hyperlink where you got this from?

The link was in my comment https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html#search-bugs. The "limit" is in the Search Bugs section table where it lists the parameters. I have not yet found anything in the API documentation to pull limit set by the administrator.

Also, you may want to see what Perceval does with their restful API endpoint. Do they ask a parameter from the user? Surely they had to also decide on this question!

It seems like Perceval is getting a max_bugs parameter 
https://github.com/chaoss/grimoirelab-perceval/blob/fd6dde67a2ce5902735d849fae7f6f68d6d85de3/perceval/backends/core/bugzillarest.py#L76-L78

And using that as the limit
https://github.com/chaoss/grimoirelab-perceval/blob/fd6dde67a2ce5902735d849fae7f6f68d6d85de3/perceval/backends/core/bugzillarest.py#L189-L190

It's an option you can use in the perceval bugzillarest command.

This appears in the -h section for perceval bugzillarest in the terminal:
```
Bugzilla REST arguments:
  --max-bugs MAX_BUGS   Maximum number of bugs requested on the same query
```

carlosparadis: > Sounds good! Also, did you see my response https://github.com/sailuh/kaiaulu/pull/177#discussion_r1162520518? 

> I didn't see an exact answer to using a different condition for the while loop, but to solve the problem where I end up writing an empty json file at the end of my download issues function, I was wondering if I could use write_json instead of htrr:write_disk? That way, I can separate the GET request from actually writing the json to a file. I tried it out and I can do it so that I don't generate an empty file at the end.

This is it: 

https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/github.R#L336

More specifically the `!is.null(gh_response)` 

With GitHub API, we request a new page with this:

https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/github.R#L344

The function used, in turn uses a library that wraps around GitHub API:

https://github.com/sailuh/kaiaulu/blob/716a46e546d5293f3b2856cace62311c95aae977/R/github.R#L290

In essence, in your code, you request the next page at every step of the loop, incrementally somewhere. Surely, at some point you will request a page + 1 that doesn't exist. You have to examine what httr returns to you when this happens, and test for it on the while loop. 

I think that's what you were trying to do with status code, but you said it won't stop the condition. Why? try running the function on a page that doesn't exist and one that exists, or maybe look how that function of gh_response is able to return a NULL value. This is likely the code logic you are missing, unless I am misunderstanding the issue.

R/download.R
```diff
@@ -12,9 +12,11 @@
 #' @param bugzilla_site URL to specific bugzilla site
 #' @param start_timestamp when to start bug retrieval (ex. 2023-01-01T00:14:57Z)
 #' @param save_folder_path the full *folder* path where the bugzilla issues will be stored
+#' @param limit the number of issues saved in each page file. Some bugzilla sites have limits set on how many bugs
+#' can be retrieved in one GET request, in which case, the limit should be set accordingly.
 #' @seealso \code{\link{download_bugzilla_comments_from_rest_api}} and \code{\link{parse_bugzilla_crawler_issue}}
 #' @export
-download_bugzilla_issues_from_rest_api <- function(bugzilla_site, start_timestamp, save_folder_path){
+download_bugzilla_issues_from_rest_api <- function(bugzilla_site, start_timestamp, save_folder_path, limit=500){
```
leilani-reich, line 19: I added a new parameter "limit" with the default value 500, since that's what Perceval's Bugzilla REST API backend used, as we discussed here https://github.com/sailuh/kaiaulu/pull/177#discussion_r1162589675.
The reason I added the limit parameter is that some Bugzilla sites limit the number of bugs that can be retrieved in a single GET request. For instance, Red Hat Bugzilla limits the bugs retrieved to 20 by default. If you use a limit > 20, it will skip over bugs. And I didn't see a way to retrieve the default limit for a Bugzilla site manually. So I made it so a user should be able to set this value.

R/download.R
```diff
@@ -23,45 +25,53 @@ download_bugzilla_issues_from_rest_api <- function(bugzilla_site, start_timestam
...
+  bug_ids <- c()
+  # Defines what bug to start from in bugs retrieved.
+  offset <- 0
+  # Defines name of the file. Each page contains 500 bugs.
+  page <- 0
+
   # Get the issues starting from the specified date
-  issues <- httr::GET(paste0(bugzilla_site, "/bug", "?creation_time=", start_timestamp, "&limit=20", "&offset=0"),
-                      httr::write_disk(file.path(paste0(save_folder_path, 0, ".json")), overwrite = TRUE))
+  issues <- httr::GET(paste0(bugzilla_site, "/bug", "?creation_time=", start_timestamp, "&limit=", limit, "&offset=", offset))
```
leilani-reich, line 36: I use write_json instead of httr::write_disk so that I can check if the json data retrieved from a GET request is empty, so I know whether I should write the json file. Before, I was writing an extra empty json file at the end (since I wasn't able to immediately tell if the JSON was empty), but separating the GET request and writing the file fixes this issue.

R/download.R
carlosparadis: > Red Hat Bugzilla limits the bugs retrieved to 20 by default. 

What do you mean by default here? If I user type limit = 500 on function call, Red Hat will accept the function call but download 20 at a time? And how did you find it was 20 by default? 

I would include this comment and the answer to my question on the function documentation body too. It sounds helpful.

R/download.R
carlosparadis: Sounds good!

carlosparadis:
This PR has conflicting files on R/parser.R and man/parse_mbox.Rd according to GitHub. You may need to Git Pull and squash to address them once you are done with the revisions.

R/download.R
leilani-reich: Hi Carlos, yes that's correct. The limit of 20 - for unauthenticated users specifically - is mentioned [here](https://communityblog.fedoraproject.org/changes-to-bugzilla-queries/#:~:text=The%20value%20of%20limit%20can,0%20as%20a%20limit%20parameter.)

So for a single GET request from Red Hat Bugzilla, only 20 bugs will be retrieved at a time. If a user was to specify a limit of 500, the first GET request would retrieve bugs 0-20, and the second would retrieve bugs 500-520, and so on, since I am using the limit to adjust the "offset" parameter, which determines which bugs to retrieve for each GET request.

carlosparadis: Interesting. One more question here, and the other I am moving to the specific part of the code.

How did you find this page? Via google? Or did their Bugzilla tracker had at the bottom of the page some link that led you there? If Bugzilla has a standard way to find these pages, it would be very helpful to who is using this function since you already went through this.

R/download.R
```diff
@@ -4,6 +4,104 @@
...
+  if (length(httr::content(issues)$bugs) > 0){
+    jsonlite::write_json(jsonlite::fromJSON(httr::content(issues, "text")),
+                         paste0(save_folder_path, page, ".json"), auto_unbox=TRUE)
+    json_issues <- rawToChar(httr::content(issues, as = "raw"))
+
+    # Get the ids of the bugs to download comments associated with these bugs in
+    # download_bugzilla_comments_from_rest_api
+    current_bug_ids <- jsonlite::fromJSON(json_issues)[["bugs"]][["id"]]
+    bug_ids <- c(bug_ids, current_bug_ids)
+    offset <- offset + limit
```
carlosparadis, line 48: My second question goes here.

> If a user was to specify a limit of 500, the first GET request would retrieve bugs 0-20, and the second would retrieve bugs 500-520, and so on, since I am using the limit to adjust the "offset" parameter, which determines which bugs to retrieve for each GET request.

Would it be correct to say that if you parsed only your first saved file to know how many issues were _actually_ retrieved, you would then be able to correct any user assumption on the limit, and safely execute the function?

I guess I am also wondering why you must rely on the "limit" and "offset" as a parameter.  With GitHub API, you are mainly iterating through pages, and there is an endpoint you can assume is equivalent to click the "Next" page to show. Is there no equivalent here?

carlosparadis (commented):
Added two questions follow-up for clarity.

R/download.R
leilani-reich: >How did you find this page? Via google? Or did their Bugzilla tracker had at the bottom of the page some link that led you there? If Bugzilla has a standard way to find these pages, it would be very helpful to who is using this function since you already went through this.

Hi Carlos, I found it on Google. I don't think there's a standard.

>Would it be correct to say that if you parsed only your first saved file to know how many issues were actually retrieved, you would then be able to correct any user assumption on the limit, and safely execute the function?

Yes, that's right. A user could wait until at least two files were created for the bugs they downloaded and check if the first json file has a limit that is under what the user specified. Then they could use the limit they see as the limit parameter. Or they could check the parsed table.

>I guess I am also wondering why you must rely on the "limit" and "offset" as a parameter. With GitHub API, you are mainly iterating through pages, and there is an endpoint you can assume is equivalent to click the "Next" page to show. Is there no equivalent here?

I have not found an equivalent. The offset & limit seems like the main means of pagination.

carlosparadis: > Yes, that's right. A user could wait until at least two files were created for the bugs they downloaded and check if the first json file has a limit that is under what the user specified. Then they could use the limit they see as the limit parameter. Or they could check the parsed table.

Actually, what I am asking is if your function can do that for the user instead. After the first page is done, could Malia parser be used to read one page of issues, assess it's length, and then use that as the `offset` thereafter? If so, the user `limit` parameter would be a `limit_upperbound`, and your function would always work.

leilani-reich: Got it, I will try that out.

leilani-reich: Yes, this works for me. Now that my function depends on Malia's, do you want me to add her parser to my pull request? Or should I wait for her parser pull request to be merged with main.

carlosparadis: Hmm, I'd say let's wait so her code goes through her and in case there are more revisions from her side! 

At this point, I'd suggest you git squash since there is some conflicts that need to be resolved before I can accept your PR. 

Also, please don't forget to expand your function body for the downloader to include the details of the limit parameter discussed here. They are very important, and I feel I will be asked this or ask it myself within a few months for forgetting! 

Also, please see the e-mail I just sent you :)

R/download.R
```diff
@@ -4,6 +4,120 @@
...
+
+    # Get the ids of the bugs to download comments associated with these bugs in
+    # download_bugzilla_comments_from_rest_api
+    current_bug_ids <- jsonlite::fromJSON(json_issues)[["bugs"]][["id"]]
+    bug_ids <- c(bug_ids, current_bug_ids)
+    page <- page + 1
+
+    # Get the table for the first page of bugs to check the number of rows
+    # and determine the actual value of the limit parameter.
+    page1_bug_table <- parse_bugzilla_crawler_issue(save_folder_path)
```
leilani-reich, line 60: Here's where I use Malia's parser to check the size of the first page of bugs. If the number of bugs in the parsed table is less than the limit_upperbound specified by the user, I will use the length of the table as the new limit for bug retrieval to make sure I don't miss any bugs.

carlosparadis: Fantastic :) Out of curiosity, does Perceval implement a similar solution, or would they skip ahead if the user sets the wrong limit? We should know this for the Wrapper functions.

leilani-reich: I think Perceval may skip ahead if the user sets the wrong limit for max_bugs https://github.com/chaoss/grimoirelab-perceval/blob/fd6dde67a2ce5902735d849fae7f6f68d6d85de3/perceval/backends/core/bugzillarest.py#L21.

I don't think I had max_bugs as a parameter for my bugzilla wrapper downloader.... I will need to add it in.

carlosparadis: I'm not sure you linked me the right part in their code. It points to an author!

It would be helpful if you could assess that, because if they skip ahead, we should warn our users on the function docs!

leilani-reich: Sorry, I updated the link above. I checked out using the --max-bugs parameter (which acts like the limit) for the Perceval Bugzilla wrapper. When I tried the --max-bugs = 500, it retrieves 20 bugs whereas with --max-bugs = 20, it retrieves 329 bugs. Thus, I can confirm that Perceval's bugzilla rest wrapper skips bugs. I will make sure to note this in the documentation for my wrapper, and add the --max-bugs parameter as well.

vignettes/depends_showcase.Rmd
```diff
@@ -96,7 +96,10 @@ project_file_network <- transform_dependencies_to_network(project_dependencies_s
                                                    weight_types = keep_dependencies_type)
 
 kable(head(project_dependencies[["edgelist"]],10))
+
 ```
+
+
```
carlosparadis, line 102: I think this is a collateral change, but since it doesn't touch the code it's alright.

carlosparadis (approved):
Ok!

carlosparadis:
Oops...not ok I see conflicts. Let me try to resolve them myself.

carlosparadis:
Checks pass affect merge conflict is resolved. So good to go.

PR #164, opened 2023-03-06
Bugzilla Wrapper Notebook

malialiu:
Showcase of bugzilla wrapper using perceval with traditional and rest API backend.
The following functions are include in the showcase:
- download_bugzilla
- parse_bugzilla
- parse_bugzillarest

PR #179, opened 2023-04-08
Kaiaulu's Bugzilla Crawler Parser + Issue_Comments Downloader/Parser

malialiu:
Here are the Bugzilla crawler parser functions. Let me know if I made any mistakes.
- parse_bugzilla_crawler_issue
- parse_bugzilla_crawler_comments

I didn't include 'issue_type' and 'issue_description' in the JSON file because I couldn't find them, and I also couldn't find 'comment_author_id.' But I did find the email of the comment author, which I extracted from the 'creator' field in the JSON file. I have put this email as the 'comment author_name' for now.


R/parser.R
```diff
@@ -4,6 +4,136 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+#' Parse Bugzilla issues data obtained from json files from Bugzilla crawler
```
carlosparadis, line 7: Use \code{\link{}} to explicitly refer to Leilani function and so when the user uses your function docs they can click and jump to it. 

In addition, use @seealso and include Leilani function there too. I have mentioned this in other revisions and I believe e-mails.

R/parser.R
```diff
@@ -4,6 +4,136 @@
...
+    }
+  }
+
+  # Rename the columns of data.table
+  setnames(result, expected_columns_names)
+
+  return(result)
+}
+
+#' Parse Bugzilla comments data obtained from json files from Bugzilla crawler
```
carlosparadis, line 84: Same concern as the first comment on function cross-referencing.

carlosparadis (changes requested):
Very very minor request for changes here. Could you attach some example outputs? 

Also, now that Leilani has the code for her downloader, could you try using her code or request her files to ensure your functions are working correctly together? I think you should do that in an R Notebook, because this is what the Bugzilla Downloader Notebook should show (and so I can run it myself).

carlosparadis:
> I didn't include 'issue_type' and 'issue_description' in the JSON file because I couldn't find them, and I also couldn't find 'comment_author_id.' But I did find the email of the comment author, which I extracted from the 'creator' field in the JSON file. I have put this email as the 'comment author_name' for now.

This is strange. @leilani-reich do you have any idea why these fields could not be found? The absence of type and description for issues are particularly worrisome, but we would like to have the comment author id too.

malialiu:
Hi Carlos,

I have just noticed that the first comment returned by "parse_bugzilla_crawler_comments" for each issue is actually the issue description. Should I make a column to put the issue description or leave it as a comment?

carlosparadis:
yes, I discussed this with @leilani-reich on our last call. But I no longer remember as to why. @leilani-reich do you remember?

leilani-reich:
@carlosparadis @malialiu For my perceval bugzilla parser functions, I made it so that the first comment, which was the issue description, would be excluded from the comment related fields in my table since it already appears in the issue related fields, like issue_description. And then I started counting comments as comments 2, 3, 4, +.

carlosparadis:
@leilani-reich Could you provide an example with an issue IRL, and maybe a screenshot like Malia did? I think what I want to be careful of here is if the actual issue description happens to appear as a first comment in the comments (meaning the API from Bugzilla is wrong, and you are correcting it on your code), or if we are making an assumption that the first comment is meant to be the issue description. An incorrect assumption here can compromise text mining analysis, so I want to make sure I understand.

leilani-reich:
For example, in this [issue](https://bugzilla.kernel.org/show_bug.cgi?id=217069), we see the image description and the first comment.
Screenshot 2023-04-21 at 10 45 47 PM

The description will not appear as a comment in comment_body, it will only appear in the issue_description field, due to how I set my bugzilla downloader functions for perceval up.

Fields for the first row for the issue with id 217069:
Screenshot 2023-04-21 at 10 48 36 PM
Screenshot 2023-04-21 at 10 49 23 PM

malialiu:
I modified my bugzilla crawler comment parser function, so it moved the first comment as issue description.

leilani-reich:
Hi Carlos, did you see my example here https://github.com/sailuh/kaiaulu/pull/179#issuecomment-1518567492 for the bugzilla issue? 

I know in https://github.com/sailuh/kaiaulu/pull/164#issuecomment-1518581298 you wanted to know whether the first comment being image description makes sense.

carlosparadis:
Reply to: https://github.com/sailuh/kaiaulu/pull/179#issuecomment-1518567492

No, sorry. It slip through. Thank you for the reminder. I still feel I am missing some information here, but the example is very helpful for me make an informed guess. You probably told me this already through other comments, but I need to see it all in one comment again so we can decide on this. Here's what I remember for the Crawler:

There is an issue endpoint, and a comment endpoint. You can download multiple issues per page per request (upper bounded by the limit imposed by the organization), but you can only download all comments per issue per request. 

1) Now, on the issue request, does the **API** provides you with a issue description? I think this is where I'm being forgetful. Did you tell me that in the issue request, the issue_description has the first comment of the issue? I am asking explicitly what the API gives you, not what you coded to be clear. 

2) Then the same question for the comment request: Does the **API** provides you with only comment body? Or does it provides you with both an issue description and a comment body?

I am trying to understand why you decided to make the comment_body API's first comment to populate the `issue_description`. Is it because that's what the issue API does? Or is there another reason? The **why** is what I am trying to understand so we can decide what makes the most sense.

leilani-reich:
Hi, I talk about your questions here https://github.com/sailuh/kaiaulu/pull/179#issuecomment-1518977862 for my Perceval bugzilla data specifically. I will double-check the bugzilla REST API data too.

carlosparadis:
Your comment above (https://github.com/sailuh/kaiaulu/pull/179#issuecomment-1519000527) redirects me to my own comment. I am a bit confused. Was it meant to the other comment I just replied with the screenshots of Perceval commands for Bugzilla?

leilani-reich:
Yes, I fixed the link. Thanks.

carlosparadis:
@malialiu Hi, yes this is OK! It should hopefully resolve once I accept the other PR in order, and you issue a merge commit here. Speaking of which, are we ready to start the PR accepts? If so, what order? did you finish the 3 asks on the notebook too? Just trying to remember where we at.

Thanks!

carlosparadis:
@malialiu Hi Malia,

The checks should now be working again (do a merge if not yet just in case too).

Please take a look on them to make sure they pass:

```
Warning: Missing link or links in documentation object 'parse_bugzilla_rest_comments.Rd':
  ‘download_bugzilla_rest_comments’
Missing link or links in documentation object 'parse_bugzilla_rest_issues.Rd':
  ‘download_bugzilla_rest_issues’
  ```
  
  Is the reason why the R-CMD-check is failing.

malialiu:
> Is the reason why the R-CMD-check is failing.
>
This seems like the reason it failed.
image

carlosparadis:
Try re-compiling the documentation so it generates the .Rd files and commit them alongside (the ones your function modified). If that doesn't fix the issue, there may be something needing to be fixed in the functions.

malialiu:
> Try re-compiling the documentation so it generates the .Rd files and commit them alongside (the ones your function modified). If that doesn't fix the issue, there may be something needing to be fixed in the functions.

It passes when I removed all the comments that link to the non-existent function.

carlosparadis:
@malialiu Hi Malia,

This merge conflict is more elaborated. Could you try to resolve the conflict just on this PR so I can try to accept it?

malialiu:
> @malialiu Hi Malia,
> 
> This merge conflict is more elaborated. Could you try to resolve the conflict just on this PR so I can try to accept it?

Yes! It should work right now.

carlosparadis (approved):
Looks good and passes the checks!

PR #197, opened 2023-04-30
Adding bugzilla wrapper functions unit tests

nicoelee123:
Test cases
download_bugzilla_perceval_traditional_issue_comments()
- Incorrect perceval path throws correct error
- Incorrect bugzilla site throws correct error
- Correct parameters return a nonempty json object

download_bugzilla_perceval_rest_issue_comments()
- Incorrect perceval path throws correct error
- Incorrect bugzilla site throws correct error
- Correct parameters return a nonempty json object

parse_bugzilla_perceval_traditional_issue_comments()
- Incorrect json input throws correct error
- Correct parameters return nonempty data.table object
- Correct parameters with comments=TRUE returns nonempty data.table object
    - ERROR: "Cannot find merge.data.table function"
 
parse_bugzilla_perceval_rest_issue_comments()
- Incorrect json input throws correct error
- Correct parameters returns nonempty data.table object
    - ERROR: I think there's an NA value somewhere in the json returned from REST api download function but not sure how to address
- Correct parameters with comments=TRUE returns nonempty data.table object
    - ERROR: Same error as before with there being NA value somewhere in json


nicoelee123:
@carlosparadis Hello, I'm just checking what are the last things that need to be done. Is resolving the conflicts all that's left? I also wanted to ask what's the best way to go about this, like is managing conflicts on the website best practice? Thank you

nicoelee123:
The commits from master are conflicting with the changes I pulled from Leilani's and Malia's branches of the functions. Since they are in master now, I think I'm going to try and reset back to before pulled their branches then just pull master and add my changes on top of that

Issue #206, opened 2023-05-05
Add Source Code Text Parser Module


carlosparadis:
While reading about [Anthony's work](https://www.peruma.me/publication/2022-rit-identifier_names/), I stumbled across [srcML](https://www.srcml.org/#home). This super simple-to-use binary offer means to scan any source code folder, file, etc, to annotate it with tags. This enables parsing any unit within files, hence opening many opportunities to perform text mining on source code.

PR #212, opened 2023-05-18
Added conf files for jhotdraw5.1, juni3.7, jrefacotry 2.6.24. 

lh-zhan:
Modified parser to parse `output_filepath` from config file to avoid duplicate output files for different projects.


carlosparadis (changes requested):
Thank you for the configs :) 

A few minor change requests were noted inline. 

In addition, I believe you forgot to change NEWS.md: 

https://github.com/sailuh/kaiaulu/blob/master/CONTRIBUTING.md#211-for-any-change-you-make

You can see an example of how they are added on the NEWS.md file itself. this case you would add it to "Minor Improvements." since it is not config only.

R/parser.R
```diff
@@ -1510,10 +1510,11 @@ parse_r_dependencies <- function(folder_path){
 #' IEEE Transactions on Software Engineering,
 #' vol. 32, no. 11, pp. 896-909, November, 2006.
 #' @export
-parse_gof_patterns <- function(pattern4_path,class_folder_path,output_filepath='/tmp/gof.xml'){
```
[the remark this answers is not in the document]
lh-zhan: oops, didn't think that's the default value due to my lack of knowledge in R, good point.

NEWS.md
```diff
@@ -11,7 +11,7 @@ __kaiaulu 0.0.0.9700 (in development)__
 
  * Bugzilla API now allows for output file to be specified. [#202](https://github.com/sailuh/kaiaulu/pull/202)
  * Paired parser functions now expects a filepath instead of a json string character. [#202](https://github.com/sailuh/kaiaulu/pull/202)
-
+ * Pass output file path into config file instead of specifying in parser.R to avoid output overriding for different projects [#206](https://github.com/sailuh/kaiaulu/pull/212)
```
carlosparadis: Oops, I see this branch still has my mistake on hyperlinking pull instead of the issue which may have confused you. The format is always linking the issue so `[#206](https://github.com/sailuh/kaiaulu/issues/206)` would be the correct way. But we can fix this once you add the next configs!

carlosparadis (changes requested):
There is just one little thing left on NEWS.md but otherwise, this is great. As noted in the inline comment, you can just fix that when you add the next batch of conf files. 

Let's keep this PR open as a way for you to keep adding conf files and allowing me to do inline code reviews on them. Once we have enough confs @rnkazman is happy with, I will merge them into the branch. 

I think we got a good standard going for the projects now 👍 .

carlosparadis:
@lh-zhan 

Hi, There is one more change request here: Please make sure to include all the information I provided you on the junit5.yml example (including the commented text): https://github.com/sailuh/kaiaulu/blob/206-add-text-module/conf/junit5.yml 

The only selection we should comment on instead of removing is the analysis section. 

I just realized the project information is entirely missing, which means we have no idea from the config file on the data provenance to download.  We also want the OpenHub URLs since we need to look at them for project selection. The uctags will also be required later on for the Class HDSM GitLog.

lh-zhan:
Hi @carlosparadis 
Nice catch! I built these files when I tested out the src_text_showcase notebook, and forgot to add the new sections in. I've just updated, please let me know if there is any change needed!

conf/jrefactory2.9.yml
```diff
@@ -0,0 +1,206 @@
...
+# information is used, see the project configuration file section at
+# the start of each R Notebook.
+#
+# Please comment unused parameters instead of deleting them for clarity.
+# If you have questions, please open a discussion:
+# https://github.com/sailuh/kaiaulu/discussions
+
+project:
+  website: https://jrefactory.soft112.com/
+  #openhub: https://www.openhub.net/p/jrefactory
```
carlosparadis, line 33: We should add this here: 

https://www.openhub.net/p/jrefactory

Are we able to roll back from J Refactory repo to jrefactory 2.6.24? Or is it an entire separate project?

conf/junit5.yml
carlosparadis, line 1: Could you also add https://www.openhub.net/p/junit to junit5 to OpenHub here? No need to add it to unit 3.7.yml since it is a separate project.

conf/jrefactory2.6.24.yml
```diff
@@ -0,0 +1,200 @@
...
+  #openhub: https://www.openhub.net/p/jrefactory
+
+version_control:
+  # Where is the git log located locally?
+  # This is the path to the .git of the project repository you are analyzing.
+  # The .git is hidden, so you can see it using `ls -a`
+
+  # jrefactory doesn't have git log, but still specify the git_log path on the config file, so the folder location can be extracted for the Notebook.
+
+  log: ../../rawdata/git_repo/jrefactory_source/.git
```
carlosparadis, line 42: We should rename this folder to indicate the version 2.6.24 as you did on jhotdraw 5.1, since we have more than one project here. It will be confusing otherwise.

conf/jrefactory2.9.yml
```diff
@@ -0,0 +1,206 @@
...
+
+version_control:
+  # Where is the git log located locally?
+  # This is the path to the .git of the project repository you are analyzing.
+  # The .git is hidden, so you can see it using `ls -a`
+
+  # jrefactory2.9 doesn't have git log, but still specify the git_log path on the config file, so the folder location can be extracted for the Notebook.
+
+  log: ../../rawdata/git_repo//newJrefactory/.git
```
carlosparadis, line 42: Instead of newJrefactory, use the actual version here too as you did with jhotdraw.

carlosparadis (changes requested):
A few more minor things I detected, but you can just send the changes once you add the new config projects. 

Primarily, let's include the openhub URL too since we are using them (and I can use their API via the config files too if I have that URL, and making sure we name the git log of projects we have more than one version with their release name rather than "new jrefactor". This is particularly dangerous since we don't have git log for some of them. 

Thanks!

carlosparadis (changes requested):
A few more minor things I detected, but you can just send the changes once you add the new config projects. 

Primarily, let's include the openhub URL too since we are using them (and I can use their API via the config files too if I have that URL, and making sure we name the git log of projects we have more than one version with their release name rather than "new jrefactor". This is particularly dangerous since we don't have git log for some of them. 

Thanks!

lh-zhan:
gotchu, I'll make sure to include these changes once I finished developing the new config files.

PR #233, opened 2023-09-28
git fake data functions, io_create_folder, io_make_sample_file, git_init

Rubegen:
…e_sample_file, git_init


R/io.R
```diff
@@ -20,3 +20,58 @@ make_temporary_file <- function(content,extension) {
...
+#' @return file with specified path and body
+#' @export
+io_make_sample_file <- function(save_filepath,body) {
+
+  error <- system2('echo',
+                   args = c("\"print('hello world!')\"",
+                            '>',
+                            file_path),
+                   stdout = TRUE,
+                   stderr = FALSE)
```
carlosparadis, line 37: This function is adding as body the `print('hello world!') to the `file_path` variable, which is no longer defined in this function!

R/io.R
```diff
@@ -20,3 +20,58 @@ make_temporary_file <- function(content,extension) {
...
+
+  # Create cmd
+  cmd <- paste("echo", shQuote(body), ">", shQuote(save_filepath))
```
carlosparadis, line 40: I didn't know of the function `shQuote`! Nice finding 👍

R/io.R
```diff
@@ -20,3 +20,58 @@ make_temporary_file <- function(content,extension) {
...
+
+  file_path <- file.path(folder_path,"hello.R")
+
+  #echo "print('hello world!')" >  path/to/folder/hello.R
+  error <- system2('echo',
+                   args = c("\"print('hello world!')\"",
+                            '>',
+                            file_path),
+                   stdout = TRUE,
+                   stderr = FALSE)
```
carlosparadis, line 64: I'm not sure what is going on here, this looks the same code as above, minus the correct command which should have been mkdir!

R/io.R
```diff
@@ -20,3 +20,58 @@ make_temporary_file <- function(content,extension) {
...
+
+  # git init path/to/folder
+  error <- system2('git',
+                   args = c('init',
+                            folder_path),
+                   stdout = TRUE,
+                   stderr = FALSE)
+
+}
```
carlosparadis, line 73: Why is git init re-defined here again?

carlosparadis (changes requested):
I am a bit confused on some of the code refactored here. Is there some difficulty understanding the command line commands and how they translate to R?

carlosparadis:
Small side note:

- Make sure to include the .Rd files too on the commit :) Those are doc files generated when you Cmd + Shift + D. 

Please do not delete the fork as you try to change the code! This will kill the Pull Request, and we don't want to create one PR to every attempt!

carlosparadis:
@Rubegen 

The last request for changes going by my comment was 4 weeks ago and your last commit 2 weeks ago. Could you give me an update here? 

Thanks!

Rubegen:
Hi Carlos, I've implemented the parameter and function changes that we discussed at the meetings, and I have been trying to install perceval on my machine to ensure the PR tests all pass. I was able to install perceval on my machine using pip3 but the tests still fail when I run them on RStudio. The 'Which Perceval' command also does not return anything and I am figuring out why.  I wanted to commit my changes to keep you updated on my code but I am still working out the perceval problems on my end.

carlosparadis:
@Rubegen Did you try to search your computer for the file "Perceval" to see if it finds anything? I believe I misremembered `which` perceval to return, but searching the computer should suffice to find the file.

Rubegen:
I found the file in my python packages. Namely /Users/rubenjacobo/Library/Python/3.8/lib/python/site-packages but I still am not able to call it. Using 'perceval' on terminal always returns 'command not found' but I am sure that I have it installed. Should I still copy and paste that directory that leads to perceval into the tools.yml page?

carlosparadis:
Can't you just run the command in the folder it is located? You don't need to be able to type "perceval" to run the tool, just have its path added on tools.yml. If it runs on its own folder it is fine.

carlosparadis:
Note you are modifying the DESCRIPTION file and the tools.yml. Neither file should be being modified by this PR. The change of version in the DESCRIPTION file may be the reason you are not passing the checks, as it is leading for the XML package to not be found. You need to reverse the change and git squash the PR to address this.

carlosparadis:
Updating your fork with most recent Kaiaulu changes, let's see if that fix your checks.

tests/testthat/test-parser.R
```diff
@@ -70,3 +70,37 @@ test_that("Calling parse_gitlog with incorrect git repo path returns correct err
...
+test_that("Git repo with 3 commits, first file is hello.R, then renamed to hi.R, then adds another file bye.R", {
+  # Create a temporary directory for the Git repository
+  tools_path <- file.path(tools_path)
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
+  git_repo_path <- suppressWarnings(example_repo_three_commits())
+  result <- parse_gitlog(perceval_path, git_repo_path)
+  # expect only 1 because prefix test files and suffix example files are ignored.
+  expect_equal(nrow(result), 3)
+  suppressWarnings(git_delete_sample_log(git_repo_path))
```
carlosparadis, line 84: You should not be suppressing warnings. If they are appearing, it is because something is wrong and you should fix it. Indeed, `git_delete_sample_log` is not deleting any folder here. Because of that, the folders don't get deleted, and the next time you run the unit test, it leads to strange errors.

R/example.R
```diff
@@ -0,0 +1,75 @@
...
+
+  # Add example_test.R file and commit it
+  test_path <- file.path(folder_path, "test-example.R")
+  io_make_sample_file(test_path, "print('tester')")
+  git_add(git_repo, folder_path, test_path)
+  git_commit(git_repo, folder_path, "Commit test-example.R file to repo", "John Doe", "JohnDoe@test.com")
+
+
+  # Add hello.R file and commit it
+  example_path <- file.path(folder_path, "fake-example.R")
```
carlosparadis, line 61: This is not the intended file format. The idea was to create two files with different prefixes, test, and example, and try to filter, resulting in the expected equal 1. Failing the test would lead to 3 files.

tests/testthat/test-parser.R
```diff
@@ -70,3 +70,37 @@ test_that("Calling parse_gitlog with incorrect git repo path returns correct err
...
+# A repo with 3 commits, where 1 file has as prefix _test.R, 1 file has the suffix example_*.R,
+# and 1 file hello.R.
+test_that("Git repo with 3 commits, first file has prefix:test, second file has suffix:example and last file normal", {
+  # Create a temporary directory for the Git repository
+  tools_path <- file.path(tools_path)
+  tool <- yaml::read_yaml(tools_path)
+  perceval_path <- tool[["perceval"]]
+  git_repo_path <- suppressWarnings(example_repo_suffix_prefix())
+  result <- parse_gitlog(perceval_path, git_repo_path)
+# expect only 1 because prefix test files and suffix example files are ignored.
```
carlosparadis, line 100: They are not ignored automatically. You still need to add filter functions.

tools.yml
```diff
@@ -1,5 +1,5 @@
 # https://github.com/chaoss/grimoirelab-perceval
-perceval: ~/perceval/bin/perceval
+perceval: ~/Users/rubenjacobo/anaconda3/bin/perceval
```
carlosparadis, line 2: You should not be editing the tools.yml. Do not use `git add .` to add everything you change. Carefully use `git diff file` to see the changes, then `git add filename` one at a time.

carlosparadis (changes requested):
The XML error (https://github.com/sailuh/kaiaulu/pull/233#issuecomment-1793338706) is addressed, however this still needs a lot of work to get closer to a merge state. 

Two functions, I imagine from @waylonho are also needed for your unit tests to work, and they are also missing here.

carlosparadis:
Your checks are now passing, and I tried to help you on the more complicated parts of the code. However, there is still work to be done on the code documentation, function naming, etc. 

You also have to carefully test the code. If you ran the git_delete_sample_log for example, you would be able to see the folder was not being deleted.

all supressWarnings should be removed and the warnings addressed. 

There are functions with missing title and incomplete descriptions.

Finally, please coordinate with @waylonho so your `git.R` are the exact same file since you both depend on the git.R file. The examples.R should only include your own functions, not the other. They should be merged after I accept either of yours into master. The same for unit tests: Only yours should be present on your commit. 

Please spend some time studying the changes and trying them out yourself and comparing to your current version to see why yours did not work and these now pass the tests. I recommend you git clone this branch of the repo in a separate folder on your computer so you can compare side by side.

Thanks.

Rubegen:
Hi Carlos, thank you for the code additions. I am currently coordinating with Waylon to ensure we're on the same page for the functions that are needed for both of our Unit tests to pass.  I looked through your comments on the code and I am working on making the proper adjustments for the prefix and suffix tests. I aim to have a new commit done by tomorrow. Just wanted to keep you posted on my work. Thank you

carlosparadis:
Thank you for staying in touch! Let me know if anything is not clear. Happy to clarify.

carlosparadis:
Forgot to mention: you both also modified and need the `io.R` file, so please also do make them consistent between both PRs (like the git.R). @waylonho notice I added a new function to delete folders on @Rubegen PR. You should use this over the git_delete_sample function where needed.

R/example.R
Rubegen: Sorry for the wrong file naming conventions, I misunderstood the issue description. So just to clarify, both "example" and "test" are prefixes for files. Neither of them are suffixes? So, some test files I could create to test my filter functions include "test-fakedata.R" and "example-fakedata.R"?

tests/testthat/test-parser.R
Rubegen: Should the filter functions be added in the git.R file?

R/example.R
carlosparadis: the suffix is the file extension, e.g. `.R` or `.java` or `.txt`. Have a look at the docs of the filter functions. In a real analysis, you don't want to be reporting the number of enhancements the `README.txt` has, but only the source code. The same is true for unit test or example files. Seldom they are the primarily object of interest when analyzing the quality of the code (unless you are explicitly interest on the unit tests quality). 

So the goal of this example was to create the 3 files, and use 2 filters. In the commit I did for you, the 2 prefix filters are utilized and correctly filtered. The suffix filter is not implemented yet, but to be honest, that is fine. I'd say, just polish the name of the variables and the docs on the example function to be a test of prefixes instead of suffix! 

There is something on this unit test I will need to change later (not due to a mistake on your part) anyway, so I'd prefer you just to make sure you understand what is going on the code, and that every line the example and unit test executes you check on the terminal it is working as intended as-is.

tests/testthat/test-parser.R
carlosparadis: On my more recent commit, I already addressed that. 

Just to be clear: The code revisions here were added to help you understand why I started making commits to fix your checks not passing. My subsequent commits addressed a number of them. 

So, check that the more recent version of my commits fixes the issue I raised, then compare to see how I addressed them. You will want to understand this well for milestone 2. 

Note I did not mention exhaustively everything you have to change on this code review. Some you will need to inspect the code yourself, and test the line execution to make sure it all work as intended. 

For example, I may not have used the new function to delete folder in all your examples, there are still functions poorly documented or without title, etc.

PR #234, opened 2023-09-29
Parallelize git log entity analysis

nicolehoess:
Adds parallelization support when analysing git log entities over multiple time windows.

Parallelization relies on the doParallel and doSNOW packages for foreach-loops.

An example implementation is shown using the Spark project configuration file.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,9 +40,7 @@ Imports:
     XML (>= 3.99-0),
     RColorBrewer (>= 1.1-2),
     cli (>= 2.0.2),
-    docopt (>= 0.7.1),
-    doParallel (>= 1.0.17),
-    doSNOW (>= 1.0.20)
+    docopt (>= 0.7.1)
 VignetteBuilder: knitr
 URL: https://github.com/sailuh/kaiaulu
 BugReports: https://github.com/sailuh/kaiaulu/issues
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -14,7 +14,8 @@ Authors@R: c(
     person('Jari van', 'Meijel', role = c('ctb')),
     person('Leilani', 'Reich', role = c('ctb')),
     person('Malia', 'Liu', role = c('ctb')),
-    person('Nicholas', 'Lee', role = c('ctb'))    
+    person('Nicholas', 'Lee', role = c('ctb')),
+    person('Nicole', 'Hoess', role = c('ctb'))
     )
 Maintainer: Carlos Paradis <cvas@hawaii.edu>
 License: MPL-2.0 | file LICENSE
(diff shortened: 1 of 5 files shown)


DESCRIPTION
```diff
@@ -39,7 +40,9 @@ Imports:
     XML (>= 3.99-0),
     RColorBrewer (>= 1.1-2),
     cli (>= 2.0.2),
-    docopt (>= 0.7.1)
+    docopt (>= 0.7.1),
+    doParallel (>= 1.0.17),
+    doSNOW (>= 1.0.20)
```
carlosparadis, line 45: Could you remove these dependencies from DESCRIPTION, and mention them on the Notebook? This still a tentative (undocumented) organization for dependencies I have been following, but basically DESCRIPTION holds dependencies that Kaiaulu API (i.e. R/*) functions depend on. Meanwhile, dependencies only a Notebook currently depends on are deferred to the user of that particular Notebook to setup. This is consistent to the philosophy of only needing to setup the third party tool of interest for the Notebook.

If too many Notebooks start relying on them, then the inconvenience is a reminder that said functionality should be transferred to within the API, rather than live on the Notebook. How well this will scale, versus try to keep up with all the tools remains to be seen. But at the very least, it has prevented me to too quickly commit to a particular library or approach, while having them still be usable :)

NEWS.md
```diff
@@ -3,6 +3,7 @@ __kaiaulu 0.0.0.9700 (in development)__
 
 ### NEW FEATURES
 
+ * A parallel version of the git log entity analysis was added to process multiple time windows in parallel. See the new `gitlog_entity_showcase_parallel.Rmd` for details. [#231](https://github.com/sailuh/kaiaulu/issues/231)
```
carlosparadis, line 6: Thank you for remembering this, a lot of people tend to forget! 👍

conf/spark.yml
```diff
@@ -0,0 +1,157 @@
...
+# If you have questions, please open a discussion:
+# https://github.com/sailuh/kaiaulu/discussions
+
+project:
+  website: https://spark.apache.org
+  openhub: https://openhub.net/p/apache-spark
+
+version_control:
+  # Where is the git log located locally?
+  log: ../../rawdata/git_repo/spark/.git
```
carlosparadis, line 37: Thank you for not using fullpaths! 👍

vignettes/gitlog_entity_showcase_parallel.Rmd
carlosparadis, line 1: The Notebook looks great! I can't test it since I am on OSX but it looks well explained on the new part! 👍

carlosparadis (changes requested):
Thank you for the PR! I only had one minor request on the use of dependencies on DESCRIPTION file I forgot to mention. Would you mind sending me another commit moving them out and adding a note on the Notebook? 

Otherwise it looks great! No need to git squash, I should be able to do directly via GitHub PR interface.

carlosparadis:
Oops! Forgot one more thing @nicolehoess. Could you also rename the notebook so it has as prefix underline? I.e.  

`gitlog_entity_showcase_parallel.Rmd` -> `_gitlog_entity_showcase_parallel.Rmd` what this will do is, it will prevent pkgdown to try and compile it when generating the itm0 docs page! :)

Thanks again!

carlosparadis (approved):
This all looks great, thank you again :)

PR #247, opened 2023-11-13
Weighted Temporal Networks and Transforms

carlosparadis:
This PR goal is to integrate weighted cumulative sum temporal networks which can use any weights assigned by the user, and also unit test them more carefully. 

A number of code parts needs to be checked in order to better improve code modularity here.


carlosparadis:
Tests are now passing, as uctags is now available on GitHub Actions. 

Some to-dos remaining is to improve the documentation of the new functions, the transform functions which use them, and the unit tests. 

Some unit tests currently generate raw data, even though they don't need it. For example test-graph.R unit tests should not rely on raw data or the parse_ functions, but rather generate the table on the fly. The raw examples should be passed on to test-parser.R and evaluate the structure of the edgelists instead.

I am also adding @nicolehoess on this assignee branch since we have already been using some of the examples to compare the behavior of edge and weght with Codeface too.

Issue #228, opened 2023-09-17
Create a fake data generator for jira issues and comments


carlosparadis:
Let's use this issue for a fake jira issue data generator. We need to first agree on the endpoint that will serve as our example data. The prior issue was to create a JIRA issue crawler, but I will recycle the issue as the endpoints need is still the same. 

Edit: Here's some more more guidance. 

# The JIRA API

At the time of this issue, JIRA has v2 and v3 (beta). Testing some endpoints of v3 against Apache JIRA, which is where most papers use data from, v2 works and v3 doesn't. So we will stick to v2. 

v2 API docs is: https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/#about

# Endpoints of interest 

After choosing the API version, what we need to select what endpoints we care about for the fake data. This has to match what the future JIRA parser should implement. Let's first consider what Kaiaulu currently uses, which is the JirAgileR package. 

## The JirAgileR Package Issue and Comment Downloader

This is the call Kaiaulu does to the package in order to download issue data:

https://github.com/sailuh/kaiaulu/blob/cb2f4098d6aeedb1cf3721a4de75803fe97145f1/vignettes/download_jira_data.Rmd#L74-L88

In turn, this is the portion of JirAgileR code that takes those parameters:

https://github.com/matbmeijer/JirAgileR/blob/7626a419f8f9e19aa6d73bb65e7a5c1c7c4da26e/R/exports.R#L540-L546

```r
  url<-httr::modify_url(url = url,
                        scheme = if(is.null(url$scheme)){"https"},
                        path = adapt_list(url$path, c("rest", "api", "latest", "search")),
                        query=list(jql=jql_query,
                                   fields=conc(fields),
                                   startAt = "0",
                                   maxResults = maxResults)
```

What we care about here is the line:

```r
path = adapt_list(url$path, c("rest", "api", "latest", "search")),
```

This means that the library Kaiaulu relies on is using the /search endpoint. On the docs, this means this endpoint is being used:

https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-search/#api-rest-api-2-search-get 

[...]

PR #251, opened 2023-11-16
i #228 refactored functions out of jira_create_sample_log

waylonho:
refactored jira_create_sample_log into create_base_info and create_ext_info functions with helper functions.

New functions currently still use the exact same data and is hard coded, will change this so parameters can be inputted. Also will write more tests. The commit serves mainly as a checkpoint.


carlosparadis:
This is a good start, but you want to be careful to, when refactoring the list, to not create an entirely different JSON. If the jira_sample_log() has a list inside a list, inside a "dictionary" for one particular field, that's because the JSON also had a list, inside a list inside a dictionary. If in refactoring your code you replace that with just a simple list, while the code looks simpler, the JSON format has changed (because the list structure has changed), and parse_jira() will fail. 

I suggest you try to just replace one value first of jira_sample_log() instead of all of them at once,call the function to save and call parse_jira to check you didn't alter the format at every small refactoring you did so you spot you broke the format before going too far and then not knowing what broke it. If it is passing, then it should be safe, but manually compare the json to the original function and the refactored one would also be a good idea. If it is passing insofar then great 👍

waylonho:
Yes, I've compared the resulting json to the original function and it looks the same, and it passes parse_jira(). I am now working on the parameters (and tests) and had a few questions:

Some fields like the components and reporter fields can be broken down and parametrized (ex. multiple components which are lists), but some other fields like resolutiondate just require one input. Which ones can have multiple (ex. can there be multiple creators?) Also, am I understanding this right? 

Can we assume the order of base_info_cell and ext_info_cell fields will always stay the same? 

I will try and have a relatively complete code by tonight for the deadline.

carlosparadis:
> Can we assume the order of base_info_cell and ext_info_cell fields will always stay the same?

Let's assume that base_info and ext_info always stay the same. These fields are not generated by JIRA, but rather the JirAgileR crawler we use, so I have no reason to believe that would keep changing. Even if they do, at some point we do plan to implement the JIRA issue Crawler anyway, so we should not worry about this. 

> Some fields like the components and reporter fields can be broken down and parametrized (ex. multiple components which are lists), but some other fields like resolutiondate just require one input. Which ones can have multiple (ex. can there be multiple creators?) Also, am I understanding this right?

This is a great question. The answer is: I have no idea. But, as you saw my e-mail, I created us our own JIRA. Please feel free to make a few issues and play around to see what can and not happen. I can delete the project later, so assume it as a sandbox. You should also try to check its online docs to see if we can find something (and please link here if you do so it is easy to find in the future!). Try to download data from our own JIRA, it may or not work. Let me know. 

https://support.atlassian.com/jira-software-cloud/docs/create-an-issue-and-a-sub-task/#Create-an-issue

Also, if you could paste me an example here where you saw multiple assignees that would be great. It doesn't seem I am able to. There are a few differences in visual appearance between ours and Apahe JIRA too, so maybe we are on different versions. 

I made some issues that reflect our own current task so you have an idea how they would look like in JIRA. Issues with checklists are called Epics. They then link to sub-issues. Have a look:

https://sailuh.atlassian.net/browse/SAILUH-1

This is one field I want to make sure we include in our fake data generator, as they offer immense data analysis potential on how an organization manages cross-issue referencing. I can't remember if I included them in our jira_sample_log. Please include this in our final fake jira data api in Kaiaulu.

carlosparadis:
@waylonho good news, it seems that we can use the API from our own Kaiaulu repo. This will give you a lot of control to understand how what you create reflects on the API. Here's how you can pull all the data:

https://sailuh.atlassian.net/rest/api/2/search?jql=project=SAILUH

However, beware: Doing so via browser uses the "actual API format on JSON". You are creating a fake data generator for the format our download_jira notebook uses instead, which uses the JirAgileR package. I unfortunately have not been able to obtain the data via there yet. Let me know if you can. 

The config file I updated for Kaiaulu, which you can change at the start of the Notebook is this (note I added the JIRA info). Don't forget to run the code blocks that says eval = FALSE manually, since by default if you do run all they will skip. 

I want to make sure we also run the download block below, in the notebook which include the field = "comment". Otherwise your sample data will not allow us to simulate comment messages and only test half the parse_jira function. Please take a look and let me know if this doesn't make sense.

```
# -*- yaml -*-
# https://github.com/sailuh/kaiaulu
#
# Copying and distribution of this file, with or without modification,
# are permitted in any medium without royalty provided the copyright
# notice and this notice are preserved.  This file is offered as-is,
# without any warranty.

# Project Configuration File #
#
# To perform analysis on open source projects, you need to manually
# collect some information from the project's website. As there is
# no standardized website format, this file serves to distill
# important data source information so it can be reused by others
# and understood by Kaiaulu.
#
# Please check https://github.com/sailuh/kaiaulu/tree/master/conf to
# see if a project configuration file already exists. Otherwise, we
# would appreciate if you share your curated file with us by sending a
# Pull Request: https://github.com/sailuh/kaiaulu/pulls
#
# Note, you do NOT need to specify this entire file to conduct analysis.
# Each R Notebook uses a different portion of this file. To know what
# information is used, see the project configuration file section at
# the start of each R Notebook.
#
# Please comment unused parameters instead of deleting them for clarity.
# If you have questions, please open a discussion:
# https://github.com/sailuh/kaiaulu/discussions

project:
  website: http://itm0.shidler.hawaii.edu/kaiaulu
  openhub: https://www.openhub.net/p/kaiaulu

version_control:
  # Where is the git log located locally?
  # This is the path to the .git of the project repository you are analyzing.
  # The .git is hidden, so you can see it using `ls -a`
  log: ../../rawdata/git_repo/kaiaulu/.git
  # From where the git log was downloaded?
  log_url: https://github.com/sailuh/kaiaulu
  # List of branches used for analysis
  branch:
    - master

mailing_list:
  # Where is the mbox located locally?
  #mbox: ../../rawdata/mbox/geronimo-dev.mbox
  # What is the domain of the chosen mailing list archive?
  #domain: http://mail-archives.apache.org/mod_mbox
  # Which lists of the domain will be used?
  #list_key:
  #  - geronimo-dev

issue_tracker:
  jira:
    # Obtained from the project's JIRA URL
    domain: https://sailuh.atlassian.net
    project_key: SAILUH
    # Download using `download_jira_data.Rmd`
    issues: ../../rawdata/issue_tracker/kaiaulu_issues.json
    issue_comments: ../../rawdata/issue_tracker/kaiaulu_issue_comments.json
  github:
    # Obtained from the project's GitHub URL
    owner: sailuh
    repo: kaiaulu
    # Download using `download_github_comments.Rmd`
    replies: ../../rawdata/github/kaiaulu

#vulnerabilities:
  # Folder path with nvd cve feeds (e.g. nvdcve-1.1-2018.json)
  # Download at: https://nvd.nist.gov/vuln/data-feeds
  #nvd_feed: rawdata/nvdfeed

# Commit message CVE or Issue Regular Expression (regex)
# See project's commit message for examples to create the regex
commit_message_id_regex:
  issue_id: \#[0-9]+
  #cve_id: ?

filter:
  keep_filepaths_ending_with:
    - R
  remove_filepaths_containing:
    - test


# Third Party Tools Configuration #
#
# See Kaiaulu's README.md for details on how to setup these tools.
tool:
  # Depends allow to parse file-file static dependencies.
  depends:
    # accepts one language at a time: cpp, java, ruby, python, pom
    # You can obtain this information on OpenHub or the project GiHub page right pane.
    code_language: java
    # Specify which types of Dependencies to keep - see the Depends tool README.md for details.
    keep_dependencies_type:
      - Cast
      - Call
      - Import
      - Return
      - Set
      - Use
      - Implement
      - ImplLink
      - Extend
      - Create
      - Throw
      - Parameter
      - Contain
  # Uctags allows finer file-file dependency parsing (e.g. functions, classes, structs)
  uctags:
    # See https://github.com/sailuh/kaiaulu/wiki/Universal-Ctags for details
    # What types of file-file dependencies should be considered? If all
    # dependencies are specified, Kaiaulu will use all of them if available.
    keep_lines_type:
      c:
        - f # function definition
      cpp:
        - c # classes
        - f # function definition
      java:
        - c # classes
        - m # methods
      python:
        - c # classes
        - f # functions
      r:
        - f # functions

# Analysis Configuration #
analysis:
  # You can specify the intervals in 2 ways: window, or enumeration
  window:
    # If using gitlog, use start_commit and end_commit. Timestamp is inferred from gitlog
    start_commit: 224a729f44f554af311ca52cf01b105ded87499b
    end_commit: 74cd4d4835a02e01e310476c6776192ad0d97173
    # Use datetime only if no gitlog is used in the analysis.
    #start_datetime: 2013-05-01 00:00:00
    #end_datetime: 2013-11-01 00:00:00
    size_days: 30
#  enumeration:
     # If using gitlog, specify the commits
#    commit:
#      - 9eae9e96f15e1f216162810cef4271a439a74223
#      - f1d2d568776b3708dd6a3077376e2331f9268b04
#      - c33a2ce74c84f0d435bfa2dd8953d132ebf7a77a
     # Use datetime only if no gitlog is used in the analysis. Timestamp is inferred from gitlog
#    datetime:
#      - 2013-05-01 00:00:00
#      - 2013-08-01 00:00:00
#      - 2013-11-01 00:00:00
```

waylonho:
Hi, I have been trying to run the notebook (download_jira_data.Rmd) and am having some trouble. I think it'd be good to hop on a call today (Monday), which I will send an email for. Currently have the functions for jira_create_sample_log (with comments and no comments), some examples written, and tests.

waylonho:
Hello,

I've replaced jira_create_sample_log with jira_create_sample_log_issues and jira_create_sample_log_issues_comments. The jira_create_sample_log_issues generates the same json as the old hardcoded log. I am able successfully call parse_jira on both of them and will commit examples and tests next.

Something I didn't think about when designing the create sample log functions was how to pass custom information for the tests. jira_create_sample_log_issues calls create_base_info and create_ext_info to make base_info_cell and ext_info_cell. create_base_info is simple and just its own function, but create_ext_info calls other helper create functions because it has much more fields. These helper create functions are the ones that can take parameters to make whatever custom details for the data. But ext_info_cell itself does not, it only calls the helper functions. If I were to call it in example.R, I can't pass parameters to create_ext_info.

The current work around is to manually create the ext_info_cell and call each helper function instead of calling create_ext_info. It seems like the most simple and not too complicated way without changing anything else, and it's just calling the helper functions in each example case:

ext_info_cell <- list(
    summary = list("Issue Description"),
    issuetype = create_issuetype(),
    components = create_components(component_details),
    creator = create_creator(),
    created = list("2007-07-08T06:07:06.000+0000"),
    description = list(),
    reporter = create_reporter(),
    resolution = create_resolution(self_url = "https://domain.org/jira/rest/api/2/resolution/1"),
    resolutiondate = list("2007-08-13T19:12:33.000+0000"),
    assignee = list(),
    updated = list("2008-05-12T08:01:39.000+0000"),
    status = create_status(),
    comment = create_comment(comment_details)
  )

R/jira.R
```diff
@@ -8,103 +8,392 @@
...
+    updated = list("2008-05-12T08:01:39.000+0000"),
+    status = create_status()
+  )
+
+  return(ext_info_cell)
+}
+
+#' Create Issue Type
+#'
+#' Create issue type cell for fake Jira issue
```
carlosparadis, line 81: Too short description again. You have to consider what this information is for, and where it is from, and guide your user here. This will require some research on your part, or experimenting with Kaiaulu issue tracker. As an example, you can see this block specify the issue type the user chose. You should, therefore, tell in this description an issue type can be a bug, an enhancement, etc and also specify `\url{https://support.atlassian.com/jira-cloud-administration/docs/what-are-issue-types/}` and `\url{https://community.atlassian.com/t5/Jira-articles/Understanding-issue-types-in-jira/ba-p/1497237}` as relevant documentation.

Basically, your function should give a bit of context to the user on how they can type other inputs. You do not need to explain all possibilities, just one (your e.g.) and reference them where they can learn more to write others on their own. Your description does need to guide the user on making sure it is consistent and ideally, enforce that via your code. For example, above, don't ask the user the id twice, where they can make an inconsistent parameter. But rather, ask once, and reinforce on your code logic they are the same information.

R/jira.R
```diff
@@ -8,103 +8,392 @@
...
+    subtask = list(subtask),
+    avatarId = list(avatarId)
+  )
+
+  return(issuetype)
+}
+
+#' Create Components
+#'
+#' Creates a list populated with components and their information from components_list
```
carlosparadis, line 114: Same as before (I will not mention this hereafter, please assume it needs to be re-worked). For components: `\url{https://support.atlassian.com/jira-software-cloud/docs/organize-work-with-components/}`. 

Also, make sure you only refer to JIRA official documentation rather than a random blog post or website.

R/jira.R
```diff
@@ -8,103 +8,392 @@
...
+#' Create Reporter
+#'
+#' Creates a reporter field with information
+#'
+#' @param self_url URL of API endpoint
+#' @param name Name of reporter
+#' @param key Key associated with reporter
+#' @param displayName Display name of reporter
+#' @param active TRUE, indicates if reporter's account is active or not
+#' @param timeZone Timezone associated with reporter
```
carlosparadis, line 176: You have to be consistent in the code style. Please make sure all your function parameters are `in_this_form` instead of `inThisForm`.

R/jira.R
```diff
@@ -8,103 +8,392 @@
...
+
+#' Create Jira Sample Log Issues with comments
+#'
+#' Creates a sample jira issue with two components WITH comments
+#'
+#' @param folder_path An optional path to where the sample .json should be created.
+#' @return The path to the sample .json file.
+#' @export
+#' @family {unittest}
+jira_create_sample_log_issues_comments <- function(folder_path="/tmp") {
```
carlosparadis, line 382: This should be (not in this form) as an example in example.R. Remember the jira_create_sample_log will go away, much as the git_sample_log. Think how the git_create_sample_log became example functions.

R/jira.R
```diff
@@ -8,103 +8,392 @@
...
+    status = create_status(),
+    comment = create_comment(comment_details)
+  )
+
+  return(ext_info_cell)
+}
+
+#' Create Comment
+#'
+#' Creates comment field with information
```
carlosparadis, line 294: https://support.atlassian.com/jira-software-cloud/docs/watch-share-and-comment-on-an-issue/

R/jira.R
```diff
@@ -8,103 +8,392 @@
...
+    )
+  }
+  return(components)
+}
+
+# each list is a component, added through create_components above
+component_details <- list(
+  list(self = "https://domain.org/jira/rest/api/2/component/12313938", id = "12313938", name = "x-core"),
+  list(self = "https://domain.org/jira/rest/api/2/component/12313939", id = "12313939", name = "x-spring")
+)
```
carlosparadis, line 135: Not sure what is going on here. Is this code living outside function declarations?

carlosparadis (changes requested):
There is still a good bit of work to go through here. You will need to make the corrections so I can do another pass, as there was too much to do in one go. 

In regards of your mention on expecting the user to define base_info and ext_info, that is a no-go: No one going to be able to do that without knowing just as much as you do about the format of the data. You need a function that encodes that logic. See, for example how @Rubegen was originally focusing on writing a function that just created a single fake e-mail, and then created another that can create a mailing list using it. 

Here, you need one function that specifies an issue passing all the parameters and using your other functions defined here. 

Ideally your various functions should be **inside** said function: No user will need to create a component on its own, they will need to create issues. Instead, the user should be presented with the minimum number of parameters they need to specify, and your create_issue_function uses this information to call your sub-functions. 

I am having you document each individual function for now as a first step, because said documentation still would need to be provided to the user when specifying the parameters anyway. There is also the question of how much you have to keep asking a user of your function to repeat themselves on every smaller function call. Let's discuss this on call today. 

Lastly, you will also need a function similar to @Rubegen mailing list, something like `make_issue_tracker` which calls your `make_issue` as a list. 

What I am trying to emphasize here is that, in addition to the inline comments, you have to think of your functions on how someone would use and make sense of them. If I tell you that I have a Fake Issue Generator, then I expect to create `issues` and subsequently simulate a `issue_tracker`. So if the user is presented with a sea of functions, they will most likely get lost on how it connects to that idea. Your abstraction has to be closer to how one would reasonably create an issue via the browser, so it follows intuition. 

You did a lot of good work here on defining the internals, but we need to do one more step to wrap that in a user interface someone can use. Please try to address the inline comments on the meantime, and we can discuss this wrapper function around it on call.

waylonho:
After looking at the feedback and thinking about the code, this is my intended plan. Make a create_issue function and a create_issue_tracker. create_issue will have the following parameters: issue_id, issue_api_url, issue_key, issuetype, status, summary, description, components, reporter, assignee. Then, create_issue takes these inputs and calls the individual functions to generate the issue format, then returns the list. The rest of the information in the smaller set fields are hard coded fake values.

create_issue_tracker takes in these lists and creates the json.

example.R would have different examples, one issue tracker with comments, one without, one to check if data is parsed properly.

carlosparadis:
> After looking at the feedback and thinking about the code, this is my intended plan. Make a create_issue function and a create_issue_tracker. create_issue will have the following parameters: issue_id, issue_api_url, issue_key, issuetype, status, summary, description, components, reporter, assignee. Then, create_issue takes these inputs and calls the individual functions to generate the issue format, then returns the list. The rest of the information in the smaller set fields are hard coded fake values.

Details below. Please remember to not change your current functions to only include the parameters below. Keep them as is (+ the changes requested inline to them). The information below below is for the new functions that will contain your existing functions.

Also some questions below in this issue. 

# 1. jira.r

##  1.1 parameters for make_jira_issue

 * issue_id: let's not ask this from the user. see or ask Ruben how I created the message id using a random function and use the same approach here.
 * issue_api_url: rename this to jira_domain_url (see https://github.com/sailuh/kaiaulu/pull/251/files#r1401068330) and ask jira_domain_url from the user. My comment mentions how you can construct issue_api_url from that and the issue_id variable.
 * issue_key: this is fine, should be a string.
 * issuetype: rename to `issue_type`, this is fine, should be a string.
 * status: this is fine, should be a string.
 * summary: this is fine, should be a string.
 * description: this is fine, should be a string.
 * components: this is fine, should be a string **but** it should separate the components by `;` (e.g. `x-core;x-spring`). Your function then use that information and place defaults to construct the rest of the information needed for your function.
 * reporter: this is fine, should be a string.
 * assignee: this is fine, should be a string.
 * comments: optional parameter that is a list (see below make_jira_issue_comment)

I am missing some parameters here that I see when creating an issue on JIRA, such as priority, linked_issues, fix_versions, labels and epic link. Were they not included in the sample_create_json?



## 1.2 parameters for make_jira_issue_comment

This function should take as parameter a character vector of strings. It then returns a list that can be passed to make_jira_issue_comments. Note you have to be able to handle 1 or more comments. The logic of how to insert that into a issue should be in make_jira_issue instead.

## 1.3 params for make_jira_issue_tracker

Let me know what you plan for this here.


# 2.example

yes that looks fine

# 3. unit tests

still pending the description on this one, may be worth considering upfront.

waylonho:
> > After looking at the feedback and thinking about the code, this is my intended plan. Make a create_issue function and a create_issue_tracker. create_issue will have the following parameters: issue_id, issue_api_url, issue_key, issuetype, status, summary, description, components, reporter, assignee. Then, create_issue takes these inputs and calls the individual functions to generate the issue format, then returns the list. The rest of the information in the smaller set fields are hard coded fake values.
> 
> Details below. Please remember to not change your current functions to only include the parameters below. Keep them as is (+ the changes requested inline to them). The information below below is for the new functions that will contain your existing functions.
> 
> Also some questions below in this issue.
> 
> # 1. jira.r
> ## 1.1 parameters for make_jira_issue
> * issue_id: let's not ask this from the user. see or ask Ruben how I created the message id using a random function and use the same approach here.
> * issue_api_url: rename this to jira_domain_url (see https://github.com/sailuh/kaiaulu/pull/251/files#r1401068330) and ask jira_domain_url from the user. My comment mentions how you can construct issue_api_url from that and the issue_id variable.
> * issue_key: this is fine, should be a string.
> * issuetype: rename to `issue_type`, this is fine, should be a string.
> * status: this is fine, should be a string.
> * summary: this is fine, should be a string.
> * description: this is fine, should be a string.
> * components: this is fine, should be a string **but** it should separate the components by `;` (e.g. `x-core;x-spring`). Your function then use that information and place defaults to construct the rest of the information needed for your function.
> * reporter: this is fine, should be a string.
> * assignee: this is fine, should be a string.
> * comments: optional parameter that is a list (see below make_jira_issue_comment)
> 
> I am missing some parameters here that I see when creating an issue on JIRA, such as priority, linked_issues, fix_versions, labels and epic link. Were they not included in the sample_create_json?
> 
> ## 1.2 parameters for make_jira_issue_comment
> This function should take as parameter a character vector of strings. It then returns a list that can be passed to make_jira_issue_comments. Note you have to be able to handle 1 or more comments. The logic of how to insert that into a issue should be in make_jira_issue instead.
> 
> ## 1.3 params for make_jira_issue_tracker
> Let me know what you plan for this here.
> 
> # 2.example
> yes that looks fine
> 
> # 3. unit tests
> still pending the description on this one, may be worth considering upfront.

Yes I see the other parameters but wasn't sure if they had to be added. They were not in the jira_create_sample_log. I guess what I'm trying to balance between is every single parameter in the fake_issues.json (self, id, name, etc., every small detail) and what the user cares about (the fields when creating an issue like Assignee, priority, etc.) Just want to make sure, should I be including ALL the fields like when I try to create an issue in the make_jira_issue function? 

For make_jira_issue_tracker, I was just planning something like Ruben's where the function would take as input a list of issues, then create the issue_tracker. Also this is a slightly silly question but is a issue tracker just a bunch of issues combined? I believe I understand what it should do but would like a little more details.

For the unit tests, I would check that:
- calling parse_jira on issue_comments with multiple comments parses all the comments
- one to check if the information for a cell is parsed correctly, not sure which specific one
- one more test and one more example situation

carlosparadis:
> Yes I see the other parameters but wasn't sure if they had to be added. They were not in the jira_create_sample_log. I guess what I'm trying to balance between is every single parameter in the fake_issues.json (self, id, name, etc., every small detail) and what the user cares about (the fields when creating an issue like Assignee, priority, etc.) Just want to make sure, should I be including ALL the fields like when I try to create an issue in the make_jira_issue function?

Ideally yes, realistically speaking, no: I still want you to graduate! So, let's just make sure the parameters of the jira_create_sample_log when someone making an issue are there. However, on your new function description, please add on the body of the function documentation that these parameters are currently not part of the implementation. 

You will also want to make sure that the parameters we are agreeing on here for parse_jira_issue() a user will type suffice for you to fill up the parameters of your current defined functions, and we are not missing anything. Did you check that?

> For make_jira_issue_tracker, I was just planning something like Ruben's where the function would take as input a list of issues, then create the issue_tracker. Also this is a slightly silly question but is a issue tracker just a bunch of issues combined? I believe I understand what it should do but would like a little more details. 

Your approach is correct, you can just group the issues together as Ruben did with e-mails. However, you must ensure when you combine them, they also match the format of you data. I believe jira_sample has two issues right? So you have to respect the format there for 2 issues when inserting 1, 2 or more. 

It's no silly question at all :) Truth be told I found my first contact with JIRA fairly overwhelming. Strictly speaking, a issue tracker, especially one like JIRA is much, much, more than just a collection of issues. However, the data Kaiaulu requests from JIRA currently is **only the issues and comments**. Therefore, for our purposes and intent at this point in time, it is reasonable to consider in Kaiaulu view of issue trackers, they are just a collection of issues.

In the long term, we may add other data requests from JIRA as we did with GitHub (e.g. Pull Requests), in which case I would extend your function to also take as parameter more than issues, maybe a second parameter as pull_request. Regardless, your code would set a good foundation for whoever extends it. 

These are all good questions, so please feel free to iterate if nothing else is clear. My main concern here is that we capture all the data from jira_sample_log, offer as little effort as possible to the function user to use them (e.g. note to create components, the user types "x-spring;x-core" instead of worrying about all the underlying structure of a component), and that you are able to use that information to call all the underlying functions you have. 

So if you want to maybe try pasting all function signatures here you have, and then the new 3 function signatures (make_jira_issue, make_jira_issue_comments, make_jira_issue_tracker), and check all parameters you defined for the 3 new functions suffice to call all the more specific functions, that can be a good sanity check too. Or you can jump to the code at this point.

waylonho:
@carlosparadis Hello. Happy Thanksgiving weekend. The most recent commit has the functions make_jira_issue and make_jira_issue_comment. I am able to call parse_jira on the both, and I believe the format matches for the comments. I haven't included all the parameters and their fields, some are still hard coded (like resolution and creator), but make_jira_issue/comment works with the given parameters. I will be committing the example.R and test-jira.R next and write the function headers properly.

carlosparadis:
@waylonho Happy Thanksgiving!

And also, thank you so much for the detailed commit message. These are very helpful for long commit changes. Because of it, I can point something that is mismatching from the specification quickly even before seeing the code:

> make_jira_issue_comment has the same parameters + comments_body.

If you look on item 1.2 of my comment above:

https://github.com/sailuh/kaiaulu/pull/251#issuecomment-1823615253

I said:

> This function should take as parameter a character vector of strings. It then returns a list that can be passed to make_jira_issue_comments. Note you have to be able to handle 1 or more comments. The logic of how to insert that into a issue should be in make_jira_issue instead.

So, you should refactor your jira_issue_comment so it only take as parameter a character vector (each element contains a body of a comment), and returns a list. This list object is then optionally passed as parameter to make_jira_issue. In this manner, your jira_make_issue is able to format a sequence of 1 or more comments into an object (do account for what it should return if the user returns an empty character array). 

It is your make_jira_issue that will contain the logic on how to insert this list of comments into the appropriate comment field. That's the intended specification.

Note it is done this way, to reduce duplicated code logic: make_jira_issue_comment in its current definition duplicate all code of make_jira_issue. So if we later find a bug there, we have to fix it on both functions. In the alternative, we only have to fix one. 

------

(*) The above is what my original recommendation was going to be, but in hindsight, an even better approach is just to move this specification make_jira_issue_comments inside the make_jira_issue, and add a comments parameter that take as input a character vector of strings, with default parameter being NULL. 

Your make_jira_issue then check if the value is NULL, it retuns a json without comments. If it does contain a character of 1 or more strings, it calls the jira_issue_comments to create the comment objects, and insert into the issue. 

The user always see a simple function interface to make issues, and issue comments, we only have 1 function to maintain, and there are no duplicated parameters needed for every issue (note you will still need the make_jira_issue_tracker).

Let me know if the specification of (*) makes sense. In practice, what this really should mean to you is that you a) delete jira_make_issue entirely, b) rename jira_make_issue_comments simply as jira_make_issue, and c) define your comments parameter as a character vector, where every element should be a character string representing a comment body, and your function tests if the user specify it or not, and if it does, it calls a function passing the character string vector to create the json format.

R/jira.R
```diff
@@ -11,14 +11,67 @@
 #' @return The path to the sample .json file.
 #' @export
 #' @family {unittest}
-make_jira_issue <- function(jira_domain_url, issue_key, issue_type_description, summary, components, description, reporter_name, assignee_name, status_description) {
+make_jira_issue <- function(jira_domain_url, issue_key, issue_type_description, summary, components, creator_name, description, reporter_name, assignee_name, status_description, comments_vector = NULL) {
```
carlosparadis, line 14: Yes, that's it! :)

waylonho:
Sorry, thought this comment posted earlier. Here are more details of the changes:

- make_jira_issue has the parameters: jira_domain_url, issue_key, issue_type_description, summary, components, creator_name, description, reporter_name, assignee_name, status_description, comments_vector = NULL
- if comments_vector is null, it generates a json without comments. Currently coding the tracker.
- There is a commented out sample test call to make_jira_issue at the bottom for reference, will move this to example.R.
- The make_jira_issue_comments function with the hard coded data looks slightly funky being in the middle of make_jira_issue, but that can be changed in the future.

waylonho:
I am having some difficulties with this make_jira_issue_tracker function, mainly the formatting part. I will hopefully have this completed and tested by the end of the day. There's a weird nesting issue that's affecting the fields in the issue tracker that's generated. I've been trying to do it manually by using for loops and calling each field as I'm not too familiar with lapply usage, but that leads to super long code. Will keep playing with it. Any tips for the lapply functions?

`make_jira_issue_tracker <- function(issues) {

  issue_tracker <- list()

  # combine the issues into a list
  issues <- list(issue1, issue2)
  issue_tracker[["base_info"]] <- lapply(issues,"[[", "base_info")
  issue_tracker[["ext_info"]] <- lapply(issues, "[[", "ext_info")

  folder_path <- "~/Desktop"
  jira_json_path <- file.path(folder_path,"fake_issue_tracker.json")
  jsonlite::write_json(issue_tracker,file.path(folder_path,"fake_issue_tracker.json"))

  issue_tracker <- list(base_info = list(), ext_info = list())

  return(jira_json_path)

}`

Here's what the structure should look like: 
d2d20f53509422f960604595b2f2412f

and here's what it currently generates:
0cb465362ed4073e77a3b497c6655644

carlosparadis:
@waylonho 

The easiest way for me to help you is if you do a minimal reproducible example. Try to make a simpler list example, and paste the code here that causes a similar behavior to the above. It will also help you troubleshoot. By the way, include the code enclosed in brackets so it is easier to read. 

```r
print("hello world")
```

Like this. :) This is also why we are using the issue tracker over e-mail exchange afterall.

Could you push with the current code version that produces the long list but is functional? And the sequence of steps that reproduces the problem on the full dataset.

waylonho:
Hi. I've committed the most recent make_jira_issue_tracker code, including a test call with two issues being passed. These are located at the bottom of Jira.R. Steps include: make issue1, make issue2, combine the two, and call make_jira_issue_tracker and pass them in. The json is generated in /tmp.

carlosparadis:
@waylonho i suggest you just try and finish this using the for loop. Make sure you address the other things we mentioned on call.

waylonho:
The most recent commit added needed documentation to the make_jira_issue function with examples. It also includes examples and test cases. For the issue tracker, I've been trying to use for loops to flatten the list structure in the json. The nested structure has been a lot harder than expected. 

Using the for loops approach, I found myself having to end up basically getting all the values from the base_info, then recreating the entire base_info structure and putting the fields back in. For ext_info, there's way more fields and would need more for loops within for loops within for loops, which I will try (along with lapply).

There were other attempts but at this point they are not feasible. Solely using lapplys does not format all the information properly as many are nested differently. Making a nested function in make_jira_issue_tracker that flattens base and ext basically leads to recreating the entire structure and filling it up, doesn't make sense to do that again after we already made the issues structure. I will call it a day and finalize this all before next meeting.

carlosparadis:
Just throwing this here as I did not have a chance to look at it yet: did you try using "[" instead of only "[[" ?

waylonho:
Hi. The most recent code has a completed make_jira_issue_tracker that handles multiple issues with and without comments. The code was somewhat long but it is able to handle and format everything using loops and fors. Additionally, there were some fields I edited to make_jira_issue to make consistent with the issue tracker format. These include assignee, creator, and reporter. There is a sample commented out call on the bottom of jira.R that shows everything needed to pass to the issue tracker.

The tests are in test-jira.R and examples in example.R. I changed the examples slightly but they still test the structure of issues and comments. The code is review ready at this point, and we will discuss on call.

carlosparadis (approved):
All tests pass on my end. Inputs were validated on call.

Issue #282, opened 2024-03-02
Refresh Capability for Downloading Github Data


Ssunoo2:
## 1. Purpose
The purpose of this issue is to create a refresh capability for the download_github_comments.rmd. Currently, the issue data is downloaded along a filepath, 100 issues per page, and does not have the capability of refreshing. Calling the downloader again will redownload all files starting from the first issue, which adds overhead and also does not allow the downloading of any issues that are beyond the reach of the API limit. 

## 2. Endpoints
On github, every pull request is an issue but not every issue is a pull request. Therefore 'issues' endpoint may return both issues and pull requests in the response. Pull requests from the issue endpoint will have an issue ID. To obtain a pull request ID, we need to use the "list pull requests" endpoint to download up to 100 issues per page. 

Further, the first message of each issue and each pull request is not considered a Github comment. 

Therefore, the original notebook relies on three endpoints: the 'Issue endpoint' to get the first comment of every issues, the 'Pull Request endpoint' to get the first comment of every pull request and the 'Issue and Pull request Comment endpoint' to get comments for the issues and pull requests together. 

Like Jira, Github has a [search endpoint:  ](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests)

### 2.1 Search by date
Github API has the ability to search by certain parameters. One of these is the ability to search by when an issue was [created or last updated: ](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-by-when-an-issue-or-pull-request-was-created-or-last-updated)
Unlike the JIRA API, this also allows us to search by hours, minutes and seconds, so the downloading of duplicate data is less likely, though not nonzero.

[Understanding the search syntax: ](https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax)
* This also specifies how to search for date ranges

The search endpoint cannot specifically download comments and so it may not be used for the comment refresh. Instead, the issue/comments endpoint has a 'since' parameter that will download all comments updated after the date. Because the 'updated_at' parameter is initialized to the 'created_at' value when the comment is created, it also downloads all comments created after this date. More information can be found [here: ](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28)

### 2.2 Search by issue range
[...]

PR #288, opened 2024-03-26
230 abstract config file parsing

anthonyjlau:
If a change is made in a configuration file, then every notebook that uses that configuration file has to be changed. In order to centralize the process and make it easier to change configuration files, the R/config.R file will store "get" functions that get the fields from the config files. These get functions can be used in place of the hard coded variable assignments. It will also make it easier to update changes to config files as only the functions in R/config.R need to be updated.


ian-lastname (commented):
Looks fine to me, documentation seems good enough for what it does.

Ssunoo2:
Nice work!

carlosparadis:
@anthonyjlau  Error occurs on my local machine too when I try to build it fails with the following error:

```
Error in parse(outFile) : 
  /Users/cvp/Desktop/kaiaulu/kaiaulu/R/config.R:770:68: unexpected symbol
769: 
770:   key <- conf[["issue_tracker"]][["bugzilla"]][["project_key_index"project_key
```

R/config.R
```diff
@@ -37,7 +37,7 @@ get_tools_project <- function(tool_name) {
 #' @export
 get_conf <- function(project_name) {
 
-  conf_path <- paste0("../conf/", project_name, ".yml")
+  conf_path <- paste0("conf/", project_name, ".yml")
```
carlosparadis, line 40: I am not sure if this function makes much sense. You will still need to hardcode the project name to the notebook, which makes sense: The notebook should work for any config file, and the user has to select one.

Remember what we want to encapsulate to prevent cascading effect is the contents of the config file itself. The path to find them is less a concern, and we certainly do not want to force the user on only allowing to choose in one place. It goes against the philosophy of letting the user type the paths of all files in the config file.

Just accept the project_name default parameter as the full path. The default value will serve as documentation for the user.

R/config.R
```diff
@@ -11,7 +11,7 @@
 #' @return the tools.yml file
```
carlosparadis, line 11: These fuctions are lacking proper title, description and parameter documentation. Refer to other files on master to see how it is done. This code will not pass the checks. This applies to all functions. Also the return tag is not quite accurate, what returns is the parsed tools.yml file.

R/config.R
```diff
@@ -55,7 +55,7 @@ get_conf <- function(project_name) {
 #' @return the local git repo path
 #' @export
 get_git_repo_path <- function(project_name) {
-  conf_path <- paste0("../conf/", project_name, ".yml")
+  conf_path <- paste0("conf/", project_name, ".yml")
```
carlosparadis, line 58: This is not what you want to do in any of the functions below. You are repeating the code over and over again. Why read the conf.yml here when you already have a get_conf? The input of these functions that access the content of a *parsed conf* should be the parsed conf itself. 

In the notebook, the user would then use get_conf("path/to/conf/project.yml"), get a conf variable which is the parsed yml, and then supply to these functions. 

The notebooks would not need to be edited other than the path to the config file, since the functions in conf.R hide the logic of how to access the elements. The notebook only needs to be changed if the path to the data is different, which makes sense to be at notebook level.

R/config.R
```diff
@@ -160,7 +160,7 @@ get_filter_commit_size <- function(project_name) {
 #' @export
 get_uctags_line_types <- function(project_name) {
 
-  conf_path <- paste0("../conf/", project_name, ".yml")
+  conf_path <- paste0("conf/", project_name, ".yml")
   conf <- yaml::read_yaml(conf_path)
 
   kinds <- conf[["tool"]][["uctags"]][["keep_lines_type"]]
```
carlosparadis, line 166: you should add a return() function at the end of every function so what is returned is explicit.

carlosparadis (changes requested):
See inline comments.

R/config.R
beydlern: Should I inform the user if the get function returns NULL?

carlosparadis: Throwing an error that the path could not be found would here, or in the others that the attribute does not exist in the config file would be helpful, yes.

R/config.R
```diff
@@ -0,0 +1,796 @@
...
+  if (is.null(tool_path)) {
+    warning("Attribute does not exist.")
+  }
+
+  return(tool_path)
+}
+
+#' Get Parsed Config File
+#'
+#' Gets the parsed configuration file for any path-specified project
```
daomcgill, line 46:
```r
#' @description This function retrieves and parses a configuration file from a specified path.
#' The configuration file is expected to be in YAML format. The function is typically
#' used in Kaiaulu projects to load settings or configurations from a config file 
#' (e.g., for tool paths, project settings, or other parameters).
```

daomcgill: Or something similar.

R/config.R
```diff
@@ -0,0 +1,796 @@
...
+  if (is.null(filter_commit_size)) {
+    warning("Attribute does not exist in the configuration file.")
+  }
+
+  return(filter_commit_size)
+}
+
+#' Get Uctags Line Types
+#'
+#' Gets the Uctags keep lines type
```
daomcgill, line 154: Also here. I do not know what this means, so I assume there are others who would also not know.

daomcgill (commented):
I made some in-line comments for some ideas on how you can make the function signatures and comments clearer. I did not do all of them, but you should retain a consistent level of detail throughout the file. Even though some code already seems simple, it is still easier to read in words.

DESCRIPTION
```diff
@@ -49,4 +49,4 @@ Imports:
 VignetteBuilder: knitr
 URL: https://github.com/sailuh/kaiaulu
 BugReports: https://github.com/sailuh/kaiaulu/issues
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.2
```
carlosparadis, line 52: Do not change this

R/config.R
```diff
@@ -0,0 +1,976 @@
...
+  tool_path <- config_file[[tool_name]]
+
+  if (is.null(tool_path)) {
+    warning("Attribute does not exist.")
+  }
+
+  return(tool_path)
+}
+
+##### Git Getter Functions #####
```
carlosparadis, line 57: Can you ensure these sections match the hierarchy shown on the new config file? For example, under tools, we would find dv8 for example. If the config specification hierarchy matches the section hierarchy, it will be much easier to find the associated get functions to that region of that config. I would even sort the function definition here to match that of the config.

R/config.R
```diff
@@ -0,0 +1,976 @@
...
+
+#' Returns the list of file extensions used for filtering files to remove.
+#'
+#' @description This function returns the list of file extensions that will be
+#' used for filtering files specified in the input parameter `config_file`. The
+#' input, `config_file` must be a parsed configuration file. The function will
+#' inform the user if the list of file extensions exists in the parsed
+#' configuration file, `config_file`.
+#'
+#' @param config_file The parsed configuration file.
```
carlosparadis, line 135: Parameters that require input parsed by other functions should always use `\code{\link{}}` please see examples in Kaiaulu codebase. Without that, a user trying to read this get function, will stare at `The parsed configuration file` and have no idea where to look for to create the parameter.

NAMESPACE
```diff
@@ -48,7 +48,49 @@ export(filter_by_file_extension)
...
+export(get_jira_keys)
+export(get_jira_project_key_name)
+export(get_keep_dependencies_type)
+export(get_mbox_archive_type)
+export(get_mbox_domain)
+export(get_mbox_key_indexes)
+export(get_mbox_mailing_list)
+export(get_mbox_path)
+export(get_nvdfeed_folder_path)
+export(get_parsed)
```
carlosparadis, line 83: This function name is too cryptic and it is not a get, it is a parse. So it should be named instead parse_project_config().

vignettes/download_mod_mbox.Rmd
carlosparadis, line 1: The notebooks documentation should be expanded to explain the region of the config file of the new specification. Please see my comment to @daomcgill on her notebook file and discuss with her. I believe this would be a task for @crepesAlot 

@crepesAlot note i still need your editable image, the image needs to be updated, and that I have not yet seen the function that creates the folder directory structure using your diagram as reference. We want said function in config.R

carlosparadis (changes requested):
Did not do an extensive review but there are a few things that needs to be changed, some minor, some will take a reasonable amount of time.

vignettes/download_mod_mbox.Rmd
daomcgill: https://github.com/sailuh/kaiaulu/pull/295#pullrequestreview-2344447754

NAMESPACE
beydlern: I use the `get_parsed()` function to parse the `tools.yml` file too.  For example, one of the notebooks configuration section would contain: 
`tool <- get_parsed("tools.yml")`
`conf <- get_parsed("conf/geronimo.yml")`

Would renaming `get_parsed()` to `parse_project_config()` and using this renamed parser function to parse the `tools.yml` and project configuration files still be accurate as a name? If not, `parse_config()` sounds more accurate.

DESCRIPTION
beydlern: Should I revert `RoxygenNote` back to `7.2.3` in my next commit?

carlosparadis: It should keep the original, yes. Each of you in the group is changing it.

carlosparadis:
Nice to see these checks passing at last, thank you for merging the master to fix on this end!

crepesAlot:
Made 1 new commit after group meeting, updated NEWS.md and a very minor change to the mailing_list.

carlosparadis:
GitHub Actions fail due to its own dependency problems. This commit passes local checks, tests and compile docs. Some notebooks were prefixed to `_` until #282 is merged as the logic of parse_github_replies was modified, and we wanted the configs to reflect the code changes there.

PR #289, opened 2024-03-28
282 GitHub downloader refresh

Ssunoo2:
The purpose of this pull request is to introduce functions to allow the github issue and comment downloader to 'refresh' by extracting the date along a filepath of the most recently 'created' issue or comment and calling the github API to download only issues/comments with 'created' values after this date


R/github.R
```diff
@@ -482,3 +583,378 @@ github_api_iterate_pages <- function(token,gh_response,save_folder_path,prefix=N
...
+#' @param date_upper_bound Optional. Specify the upper bound date time (e.g. 2023/11/16 21:00)
+#' @param verbose boolean value. When set to true, it prints operational messages including
+#' greatest dates and the file name that contains the greatest date.
+#' @export
+#' @seealso  \code{link{github_api_project_issue_or_pr_comment_refresh}} to refresh comment data
+#' @seealso  \code{link{github_api_project_issue_or_pr_comments}} to refresh issue data
+github_api_project_issue_by_date <- function(owner,
+                                             repo,
+                                             token,
+                                             date_lower_bound = NULL,
```
anthonyjlau, line 918: Modify parameter name to date_created_lower_bound

R/github.R
```diff
@@ -482,3 +583,378 @@ github_api_iterate_pages <- function(token,gh_response,save_folder_path,prefix=N
...
+         since=since,  # Pass the `since` parameter in the API request
+         page=1,
+         per_page=100,
+         .token=token)
+  return(gh_response)
+}
+
+#' Download Github Issue Data by Date
+#'
+#' Appends a 'created' field to a search github JQL query and returns the first page of the response.
```
anthonyjlau, line 890: Make it clear that the function is using the created date rather than the updated date endpoint parameter.

carlosparadis:
I've modified the Notebook to simplify the narrative, since the old endpoint and new endpoint was mixed in. It should be easier to also debug now.

In addition to the issue I raised with the refresher dependency on two folders, one of which I thought it should not be going after (the issue folder), the search endpoint refresher is not working for me. You will see the Notebook tries to download early project issue data, and subsequently it should refresh it to the more current date from there.

The refresher does not add issue data. I suspect looking at the existing issue folder may be the culprit, as it does not follow the file name convention.

Ssunoo2:
The Issue refresher no longer takes in the file path for the issue data. Previously, if the issue folder was empty, it would download all data from the issue endpoint into the issue folder. Now, it only checks the issue_search folder and when empty will download all data from the search endpoint into the issue_search folder

carlosparadis:
@beydlern just noticed the review request. Is this good to go on review?

beydlern:
@carlosparadis Yes, three notebooks were refactored to expect the use of the getter functions in R/config.R in [my recent commit](https://github.com/sailuh/kaiaulu/pull/289/commits/5e90f3daf5f1028cf1e5715db7ac51a27b098c40), which are ready for review.

carlosparadis (approved):
All documentation can now be generated including social smells, motifs etc on the new configuration format. GitHub Actions is broken again, but tests and checks pass local.

PR #295, opened 2024-04-19
Created parse_mbox_latest_date and refresh_mbox functions and updated…

ian-lastname:
… helix config in accordance to new save file structure

I have created the parse_mbox_latest_date and refresh_mbox functions. The latter function deletes the latest year and month mbox file that is currently downloaded (identified by parse_mbox_latest_date), and redownloads that along with any file beyond up until the current year. The naming convention of the downloaded files are also changed to what we have agreed on. Just to note, download_mod_mbox REMAINS UNCHANGED since I'm only using download_mod_mbox_per_month.


carlosparadis:
Thank you @ian-lastname. I will try to make a pass before our meeting!

carlosparadis:
@ian-lastname thanks!

daomcgill:
Hi @carlosparadis,

I've refactored the download_pipermail() function. 

#### Proposed Changes
- Removed archive_url and archive_type parameters to simplify the function interface.
- Added start_year_month and end_year_month parameters for date filtering.
- Removed convert_pipermail_to_mbox(), since download_pipermail() handles file conversion automatically.
- Changed the file naming convention to 'kaiaulu_'YYYYMM.mbox' for consistency.

I've added temporary configuration entries in helix.yml for testing purposes:

```r
conf <- yaml::read_yaml("conf/helix.yml")
mailing_list <- conf[["mailing_list"]][["mod_mbox"]][["pipermail_key"]][["mailing_list"]]
start_year_month <- conf[["mailing_list"]][["mod_mbox"]][["pipermail_key"]][["start_year_month"]]
end_year_month <- conf[["mailing_list"]][["mod_mbox"]][["pipermail_key"]][["end_year_month"]]
save_folder_path <- conf[["mailing_list"]][["mod_mbox"]][["pipermail_key"]][["save_folder_path"]]
```
And this function call:
```r
download_pipermail(
  mailing_list = mailing_list,
  start_year_month = start_year_month,
  end_year_month = end_year_month,
  save_folder_path = save_folder_path
)
```
#### Testing Results
- Retrieves links according to date range
- Downloads and renames files from links
- Unzips and cleans gz files
- Saves files as mbox with new naming convention
- Note: I attempted to download and decompress the files directly without saving the .gz to disk, but could not establish a valid connection. The current implementation still downloads the gzipped file and then removes it after unzipping.

carlosparadis:
@daomcgill I made an inline comment to reply to your question of changes, since there may be a bit of misunderstanding. Let me know if you can't find it. 

Two sanity checks:

- Did you try running parse_mbox() to see if it works on the renamed files yet?
- Does running the refresh function pointed at a empty folder downloads the files up to date?
- Does deleting the most recent mbox file download only the newest file back?
- Does running the refresh without deleting any file also update the more recent file only? (to add more recent day's worth of data to the current file)
- Suppose you ran the current refresh today, Sep 16. Then also supposed the next time you executed again the refresh was Oct 2. Would it overwrite the Sep file, and then download a Oct file?

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
-  #Extract href tablenodes from html table
-  tableNodes <- XML::getNodeSet(tbls_xml, "//td/a[@href]")
+  # Create directory if it does not exist
+  if (!dir.exists(save_folder_path)) {
+    dir.create(save_folder_path, recursive = TRUE)
+  }
 
-  #Extract filenames from tablenode content with xmlGetAtrr
-  hrefs <- sapply(tableNodes, XML::xmlGetAttr, 'href')
+  # Ensure mailing_list URL ends with a slash
```
carlosparadis, line 23: Could you add to the comment why we need to ensure the "/"?

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+  if (!stringi::stri_endswith_fixed(mailing_list, "/")) {
+    mailing_list <- paste0(mailing_list, "/")
```
carlosparadis, line 25: For consistency, you may want to use stringi::stri_c instead of paste0

R/mail.R
```diff
@@ -6,110 +6,249 @@
 
 ############## Downloader ##############
 
-#' Download all pipermail files in an archive
-#' @param url An url pointing to a pipermail archive
-#' @return Returns `destination`, a vector of the downloaded files in the current working directory
+#' Download all pipermail files in an archive as mbox files
+#' @param mailing_list The name of the mailing list being downloaded
```
carlosparadis, line 10: Functions can have descriptions (and they can be fairly long). I would like you to add to the description of the function that it will download a .html file containing the actual URLs for the files, and subsequently download the files in .txt or .gz to convert to .mbox here. See other functions in git.R such as parse_gitlog() for examples.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
-
-  #Parse html file into object
-  tbls_xml <- XML::htmlParse(pagedata)
-
-  #Extract href tablenodes from html table
-  tableNodes <- XML::getNodeSet(tbls_xml, "//td/a[@href]")
+  # Create directory if it does not exist
+  if (!dir.exists(save_folder_path)) {
+    dir.create(save_folder_path, recursive = TRUE)
+  }
```
carlosparadis, line 21: Does this assume only the ending folder is not created? If so, I suggest you remove this code. One of the functions I will ask the config.R folks is to create a function that defines all the folder structure in Kaiaulu for the user to avoid this. So rather than us adding logic in every function for the ending folder in a path creation, the convention is enforced in a single place, the config.R, where all paths are also specified in the config.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+  # Table rows
+  rows <- XML::getNodeSet(doc_obj, "//tr")
+
+  # Skip header row
+  data_rows <- rows[-1]
+
+  # Vector for link storage
+  links <- c()
+
+  # Extract the date and link from each row
```
carlosparadis, line 47: For this and other comments, it would be a good idea to be a bit more verbose. For example, you could say the "date and link from each row, which represent a month mail file" etc. R is not the most readable language, and since this is code other students will be looking yet, it is ok to be excessive.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+    date_parsed <- as.Date(paste0("01 ", date_cleaned), format = "%d %B %Y")
+    if (is.na(date_parsed)) {
+      warning("Date could not be parsed: ", date_cleaned)
+      next
+    }
+    year_month <- format(date_parsed, "%Y%m")
+
+    # Check if date is within range
+    if (year_month >= start_year_month & year_month <= end_year_month) {
+      # Get href from column 3
```
carlosparadis, line 63: what is contained in column 3?

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
-  #Extract filenames from tablenode content with xmlGetAtrr
-  hrefs <- sapply(tableNodes, XML::xmlGetAttr, 'href')
+  # Ensure mailing_list URL ends with a slash
+  if (!stringi::stri_endswith_fixed(mailing_list, "/")) {
+    mailing_list <- paste0(mailing_list, "/")
+  }
 
-  #Create Vector
-  files <- vector()
+  # Get mailing list contents
```
carlosparadis, line 28: I think this is the place where the file with all the mail urls are displayed? I would be more verbose here, since that starts the first part of the code logic to scan the file.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+      # Store the link in links
+      link <- link_nodes[1]
+      links <- c(links, link)
     }
   }
 
-  destination <- vector()
-  # File downloading loop
-  for (i in files){
+  # Vector for downloaded files
```
carlosparadis, line 75: I would add a more explicit division in comments, maybe ####### Use Links to download individual files #### or something of the sort (and I would also add something equivalent at the first part of the code). Your code flow has two moments, one to download the first file and extract the URLs out of it to know what is possible to download, and a second where you use that information to download exact 1 mail file (either txt or gz). We want to make that logic more explicit through the comments too. Other downloaders do not follow this conceptual idea.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+      warning("Could not parse date from link: ", link)
+      next
+    }
+    year_month_clean <- format(date_parsed, "%Y%m")
+
+    # Download URL
+    txt_url <- paste0(mailing_list, gsub("\\.gz$", "", link))
+    gz_url <- paste0(mailing_list, link)
+
+    # Attempt to download the .txt file first
```
carlosparadis, line 95: good comment!

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+    download_url <- txt_url
+    response <- httr::GET(download_url, httr::timeout(60))
+
+    if (httr::status_code(response) != 200) {
+      # Fallback to .gz file if .txt is unavailable
```
carlosparadis, line 100: good comment!

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+      download_url <- gz_url
+      response <- httr::GET(download_url, httr::timeout(60))
+      if (httr::status_code(response) != 200) {
+        cat("Both .txt and .gz downloads failed for link: ", link, "\n")
```
carlosparadis, line 104: Should this be a warning() instead of a cat? stop() will break flow, but I believe warning will call more attention without stopping the code.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+    }
 
-  #Return filenames
-  return(destination)
+    # Define the destination file
+    dest <- file.path(save_folder_path, paste0('kaiaulu_', year_month_clean, '.mbox'))
+
+    # Print diagnostic info
+    cat("Downloading: ", download_url, "\n")
+    cat("Saving to: ", dest, "\n")
```
carlosparadis, line 114: Add a verbose parameter to the function, and specif these under "if verbose == true". Any prints should be optional.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+  # Return downloaded files
+  return(downloaded_files)
 }
 
 
-#' Convert pipermail archive files (.txt and .txt.gz) into an mbox format for use with \code{\link{parse_mbox}}
-#' @param filelist A vector of pipermail archive files from \code{\link{download_pipermail}}
-#' @return Returns `output`, the name of the resulting .mbox file in the current working directory
+#' Refresh mbox files downloaded via pipermail
+#' Uses the adopted file name convention by \code{\link{download_pipermail}} to identify
```
carlosparadis, line 150: Add one empty line here with `#'` to separate title and description. You may also want to double check the actual documentation of the function in Rstudio after running build (on osx cmd + shift + b) to see it is formatted poperly when the user call the function docs `?refresh_pipermail`.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
 
-    rstring <- sub(" at ", "@", string)
+  # Check if the folder is empty
+  files_in_folder <- list.files(save_folder_path, pattern = "kaiaulu_\\d{6}\\.mbox$")
+  if (length(files_in_folder) == 0) {
+    # If empty, download from start_year_month to the current month
+    end_year_month <- format(Sys.Date(), "%Y%m")
+    cat("Folder is empty. Downloading from", start_year_month, "to", end_year_month, "\n")
+    download_pipermail(mailing_list, start_year_month, end_year_month, save_folder_path)
+    return(NULL)
```
carlosparadis, line 173: Could we add an else here instead of return(NULL)? would make the code flow more clear.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+  # Redownload from the most recent month to the current real-life month
+  end_year_month <- format(Sys.Date(), "%Y%m")
+  cat("Redownloading from", recent_month, "to", end_year_month, "\n")
+  download_pipermail(mailing_list, recent_month, end_year_month, save_folder_path)
+}
 
-  }
 
-  output <- "output.mbox"
+#' Process .gz files in a folder, unzip and convert them to .mbox
```
carlosparadis, line 193: Missing title. May want to check other function docs in this PR to make sure they have the proper format.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+#' Checks a folder for any .gz files, unzips them, and renames them
+#' to .mbox format. The original .gz files are deleted after unzipping. If a .mbox
+#' file with the same name already exists, it will be overwritten.
+#'
+#' @param folder_path The path to the folder containing both .gz and .mbox files.
+#' @return A list of the .mbox files that were created or updated.
+#' @export
+process_gz_to_mbox_in_folder <- function(folder_path) {
```
carlosparadis, line 201: this function is not currently called on the refresh_ function. It should be called on every refresh call to ensure the ending state of the folder on refresh is always preserved with .mbox. I would not add that logic to download_pipermail(), in case the user wants to have the original files, they can use download_pipermail() directly.

R/mail.R
```diff
@@ -6,110 +6,249 @@
...
+  }
 
-    #Open read connection
-    readCon <- file(filename, "r")
+  # Vector to store names of converted .mbox files
+  converted_mbox_files <- c()
 
-    data <- readLines(filename)
+  # Process .gz files
+  for (gz_file in gz_files) {
```
carlosparadis, line 219: Is the logic of converting gz files to mbox duplicated in download_pipermail? If so I guess it is ok just want to make sure.

carlosparadis (changes requested):
@daomcgill i've added some inline comments, for sanity sake you can reply to each comment directly in line on these since they are more specific.

R/mail.R
daomcgill: It is duplicated, which is why I was wondering if we need this function.

R/mail.R
```diff
@@ -21,11 +21,11 @@
 #' @param mailing_list The name of the mailing list being downloaded (e.g. "https://mta.openssl.org/pipermail/openssl-announce/")
 #' @param start_year_month The year and month of the first file to be downloaded (format: 'YYYYMM')
 #' @param end_year_month The year and month of the last file to be downloaded (format: 'YYYYMM', or use 'format(Sys.Date(), "%Y%m")' for the current month)
-#' @param save_folder_path The folder path in which all the downloaded pipermail files will be stored
+#' @param mbox The folder path in which all the downloaded pipermail files will be stored
```
carlosparadis, line 24: I believe you want to keep at least `folder_path` here. The `mbox` name does not make it clear to me a path to anything is required, nor a folder. Unclear why the configs are being changed either, it is breaking the consistency.

vignettes/parallelized_parse_mbox.Rmd
carlosparadis, line 1: This is not part of this PR. Keep it in a separate branch and PR tied to the other issue. I won't be able to merge the code if we add all your milestones in this PR. Moreover. the Parallel will extend beyond mbox.

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,182 @@
...
+  require(XML)
+  require(httr)
+```
+
+
+# Introduction
+
+Mailing list data is stored in a variety of archives. See:
+- Mod Mbox: [Apache Geronimo](https://geronimo.apache.org/mailing-lists.html)).
+- Pipermail: [OpenSSL](https://mta.openssl.org/mailman/listinfo/).
```
carlosparadis, line 33: Add a comment here that you will expand on the explanation below, I almost thought you meant to only provided this as background.

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,182 @@
...
+# Pipermail
+
+## Mailing List Organization
+
+Mailing lists are typically organized by topic or purpose. For example, the [OpenSSL project](https://www.openssl.org/community/mailinglists.html) maintains several mailing lists, each serving a different group:
+
+- **openssl-announce**: For important announcements.
+- **openssl-commits**: For commit messages.
+- **openssl-project**: For project discussions.
+- **openssl-users**: For general user questions and discussions.
```
carlosparadis, line 45: This is good, but i'd rather you are not specific, instead say "project-announce" etc. It is ok on the top you provide the direct URLs as examples however.

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,182 @@
...
+
+Each mailing list maintains archives of past messages, often organized by month and year. These archives can be accessed and downloaded for analysis.
```
carlosparadis, line 47: This should be expanded. In addition to having mailing lists as different topics, each mailing list history may be split into multiple archives. See OpenSSL and create a example out of that, as I did with you on call.

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,182 @@
...
+
+```{r}
+conf <- yaml::read_yaml("../conf/helix.yml")
+mailing_list <- conf[["mailing_list"]][["pipermail"]][["project_key_1"]][["mailing_list"]]
+start_year_month <- conf[["mailing_list"]][["pipermail"]][["project_key_1"]][["start_year_month"]]
+end_year_month <- conf[["mailing_list"]][["pipermail"]][["project_key_1"]][["end_year_month"]]
+mbox <- conf[["mailing_list"]][["pipermail"]][["project_key_1"]][["mbox"]]
+```
+
+### Explanation of Configuration Parameters
```
carlosparadis, line 61: No need for a new section here.

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,182 @@
...
+- mailing_list: The URL of the mailing list archive index page (e.g., https://lists.openssl.org/pipermail/openssl-users/).
+- start_year_month: The starting date for downloading archives (in YYYYMM format).
+- end_year_month: The ending date for downloading archives (in YYYYMM format).
+- mbox: The local directory where the downloaded archives will be saved.
+
```
carlosparadis, line 66: This does not capitalize on your explanation of `##Mailing List Organization`. This section is where you tell the user how mailing lists that have multiple topics and multiple archives are organized in a distilled project configuration file. You want to add a code block here with the actual config file region where the information is provided, so the user can understand how they typing information there gets translated to the loaded strings.

This is very important, because it is on the Notebook that requires the functionality the user seeks, that the explanation of the project configuration file needed parameters happen. Contrast that to teaching the user all of the project config file for parameters they will not use, and that means this Notebook section is the best place to teach them. 

You want something like this:

https://github.com/sailuh/kaiaulu/blob/c7811069ef3c7943b509c76d4b9116d1a14dc343/vignettes/gitlog_showcase.Rmd#L35-L81 

See for example how a user would approach Kaiaulu here: https://github.com/sailuh/kaiaulu/discussions/311#discussion-7265511

vignettes/download_mod_mbox.Rmd
carlosparadis, line 1: Is there a reason this file was preserved? download_mail.Rmd is the way to go, one notebook to explain how download mailing lists is done in Kaiaulu, and explaining how to use the functions available and the concept of refresh.

carlosparadis (changes requested):
More revisions and a few questions.

vignettes/download_mail.Rmd
daomcgill: Can you please expand on this?

carlosparadis: In your OpenSSL mention, if you access the link in your notebook, you will see that the archive of a mailing list, for example -dev, may be incomplete. I believe the website even mentions one archive file is more complete than other. In short: A user may need to carefully examine each archive file rather than blindly choose any and go with it. Does that make sense? The choice of archive is contrained by Kaiaulu being able to parse it (pipermail / mod_mbox / something else in the future), and the time coverage. Sometimes it is more than time that is lacking, maybe one archive type will contain the "in_reply_to" information, whereas others may not. This should be cautioned on your notebook.

daomcgill: Thank you for clarifying!

carlosparadis:
@daomcgill one of @rnkazman students is needing this to download mod_mbox for Apache Maven. The Notebook documentation is already there right? I know it is functional from the last demo.

daomcgill:
@carlosparadis the updates are finalized and awaiting your review.  @rnkazman please let me know if your student runs into any issues using it for Apache Mavern.

carlosparadis:
@daomcgill thanks! It's been on my to-do list, on the code side but I was happy to see the refresh was in order from our last group call. You did get the 10 page long e-mail of M3 right?

daomcgill:
@carlosparadis I received it, thank you. I am working on figuring out where to start, and should have an issue up for it soon.

carlosparadis:
@daomcgill no worries. Remember, you can always start from a semi empty issue and we can q&a out of it on it.

carlosparadis:
@daomcgill 

A couple things: 

- Narrative of the notebook still needs some massage. See the text I added at the beginning as an starting point. Move the config file to the top and brief the user in one go from there instead of showing in parts for the different archives. It should be relatively clear from the name in the new config they are separate.

- New config specification and gets are still missing. I recommend at least use the new config format without the get, so that you can write the narrative with the proper explanation so I can review.

- Many of my commits above is trying to resolve why Actions is failing. Take a look at the commit messages to see what I tried. You will need to revert to OS 13 (i.e. see what I changed on the last commit and undo). From there, you will need to figure out why the gcc compiler is not working on Actions anymore. 

This used to work, but it is not the first time Actions stop working. Moving to 14 is a no-go because of another issue with uctags, a separate dependency of Kaiaulu that I would like us to still be able to test. 

The good news is that the checks passed on my machine, but we can't have my offline computer being the only way for your group to run Checks and unit tests.

tools.yml
```diff
@@ -7,7 +7,7 @@ refactoring_miner: ~/RefactoringMiner-1.0/bin/RefactoringMiner
 # https://github.com/boyter/scc
 scc: ~/scc/scc
 # universal-ctags
-utags: /usr/local/Cellar/universal-ctags/HEAD-62f0144/bin/ctags
```
carlosparadis, line 10: have to be careful with these. GitHub Actions use them. This could be the source of some checks failing.

carlosparadis (changes requested):
# Notrebook

This is too formulaic:

> Description:
> - start_year_month: The starting date for downloading archives (in YYYYMM format).
> How it works: etc

This level of detail (“what do I type in this parameter”?) should reside on the function docs (I moved some so you have as example). Also, the config does not have the start_year_month parameter, but it is described in the notebook.  See how I edited the configuration file. I moved most of them to inline on the config file itself, and focused on the manual aspect the user will need. It is easier for the user to see what they have to type where they have to type. This, in turn, will appear on your quoted docs in the notebook.

Use the blog post analogy: You would not want to read a blog post that keeps stating “Description:” and going in a lot of specifics while you are trying to get the bigger picture. You want to focus on the narrative, the analysis, and place the specifics closer to where they will be needed (the config file, the function docs, or the inside of the function). Different level of detail for different people with different purposes.

> - save_folder_path: The local directory where the downloaded archives will be saved (if you run the code in this notebook, the archives will be saved in a folder 'extdata', located in the parent directory of kaiaulu (wherever your kaiaulu folder is kept)).

I am not sure what extdata is. Should it not save to save_folder_path?


# Code blocks flying around 

While I asked for the explanation of the config to be moved together to the start, I didn’t mean to also move the code blocks needed to run your notebook. E.g.


```{r}
conf <- parse_config("../conf/helix.yml")
mailing_list <- get_pipermail_domain(conf, "project_key_1")
start_year_month <- 202310
end_year_month <- 202405
save_folder_path <- get_pipermail_path(conf, "project_key_1")
```

And 

```{r eval=FALSE}
conf <- parse_config("../conf/helix.yml")
mailing_list <- get_mbox_domain(conf, "project_key_1")
start_year_month <- 202310
end_year_month <- 202405
save_folder_path <- get_mbox_path(conf, "project_key_1")
```

Share the same variables, and therefore, will overwrite each other. You should keep the same variables, but move these code blocks to their respective downloader sections.

# Config files

Whatever is not commented has to make sense. I see on helix savepaths pointing to Kaiaulu (??). 


# Refresh

Config specifies a file, but your refresher points to multiple files. How it parses multiple files? Should it not be the folder path?

carlosparadis:
This is one of the most 4 used notebooks in Kaiaulu, so I will need to be picky about the narrative. Know a few users who will drown in the current detail. I also think the separation of where documentation should go is very important, especially now you are carrying the torch on M3 in the data science side. Please give another try.

carlosparadis:
Looking at these errors, I suspect the issue lies on Kaiaulu not being able to create a sample git log to run its unit tests in this PR. One of the errors of GitHub Actions points to line 34 of test-git.R:

https://github.com/sailuh/kaiaulu/blob/513a3f08514421160a0f58d5b7461fa0d870169e/tests/testthat/test-git.R#L34

The line right above creates the sample file:

https://github.com/sailuh/kaiaulu/blob/513a3f08514421160a0f58d5b7461fa0d870169e/tests/testthat/test-git.R#L33

And this is what the function does, i.e create a sample log on /tmp/:

https://github.com/sailuh/kaiaulu/blob/513a3f08514421160a0f58d5b7461fa0d870169e/R/git.R#L668-L686

The mystery, assuming I am on the right trail, is why it can't here but it can in the other PR.

carlosparadis:
Also this error on the actions log is what led me down this path:

>  ── Error ('test-src.R:11:3'): filters can be used to delete unit tests and example files without deleting source code ──
> Error in `parse_gitlog(perceval_path, git_repo_path)`: The repository specified has no commits.

We don't git clone anything on the bash code that runs on GitHub Actions side. It uses Kaiaulu to fake a git log (there is a whole cheat sheet of the group who made this versioned in the cheat sheet repo if you're curious how it works).

> ── Error ('test-mail.R:31:3'): Calling parse_mbox with correct perceval and mbox path returns a data table with correct > raw data ──
> Error in `.checkTypos(e, names_x)`: Object 'reply_from' not found amongst []

This error is a bit misleading because we don't have a try catch in the function, but it can't find said column because the table is empty. Again suggesting in this branch Kaiaulu can't use the fake data generators, leading to all tests failing (can't test what doesn't exist!). 

> ── Warning ('test-mail.R:15:3'): Incorrect mbox path to parse_mbox returns empty table ──

Again and again the issue persists. The fake data generator does not need a third party tool either, it is all written in Kaiaulu. The git fake data generator does need the git commands, but given mbox is also failing, I doubt git itself is the issue on Actions side. 

In short, I suspect something in this PR is hurting the fake data generator part of Kaiaulu, and the unit tests are reminding us to not proceed (unless there is something even more obscure going on and giving the false impression).

R/mail.R
```diff
@@ -4,238 +4,566 @@
...
+    if (verbose) {
+      cat("No files found in save_folder_path\n")
+    }
+  }
 
-  #Loop through time and compose the mbox file
-  for (year in (from_year:to_year)) {
+  if (length(missing_months) == 0) {
+    if (verbose) {
+      cat("No missing months\n")
```
carlosparadis, line 491: Any reason cat is being used here? I believe Kaiaulu uses print everywhere. Should change for consistency.

_pkgdown.yml
carlosparadis, line 1: Thank you for accounting for this.

conf/helix.yml
```diff
@@ -48,26 +48,31 @@ version_control:
...
+      save_folder_path: ../../rawdata/kaiaulu/mod_mbox/save_mbox_mail_2/
+      # mbox_file_path is for use only with parse_mbox() function. It is the file to parse
+      mbox_file_path: ../../rawdata/kaiaulu/mod_mbox/save_mbox_mail_2/kaiaulu.mbox
+  # If project uses Pipermail
+  pipermail:
+    project_key_1:
+      mailing_list: https://mta.openssl.org/pipermail/openssl-users/
+      save_folder_path: ../../rawdata/kaiaulu/pipermail/save_mbox_mail/
+      # mbox_file_path is for use only with parse_mbox() function. It is the file to parse
+      mbox_file_path: ../../rawdata/kaiaulu/pipermail/save_mbox_mail/kaiaulu.mbox
```
carlosparadis, line 70: kaiaulu on helix uncommented is confusing, should either comment or fix

exec/mailinglist.R
```diff
@@ -51,56 +47,62 @@ if(arguments[["tabulate"]] & arguments[["help"]]){
...
+    save_folder_path = save_folder_path,
+    verbose = TRUE
+  )
+
+  cli::cli_alert_success(paste0("Downloaded mailing list archives were saved at: ", save_folder_path))
+
+} else if (arguments[["-h"]] || arguments[["--help"]]) {
+  cli::cli_alert_info(doc)
+} else if (arguments[["--version"]]) {
+  cli::cli_alert_info('Kaiaulu 0.0.0.9600')
```
carlosparadis, line 104: 9700

exec/mailinglist.R
```diff
@@ -51,56 +47,62 @@ if(arguments[["tabulate"]] & arguments[["help"]]){
...
-  mbox <- download_mod_mbox_per_month(base_url = mod_mbox_url,
-                            mailing_list = mailing_list,
-                            from_year=start_year,
-                            to_year=end_year,
-                            save_folder_path = save_path,
-                            verbose = TRUE)
-
-  cli_alert_success(paste0("Downloaded mailing list was saved at: ",save_path))
+  conf <- yaml::read_yaml(conf_path)
+  mailing_list <- conf[["mailing_list"]][["pipermail"]][["project_key_1"]][["mailing_list"]]
```
carlosparadis, line 89: should also use get @crepesAlot  @beydlern

exec/mailinglist.R
```diff
@@ -51,56 +47,62 @@ if(arguments[["tabulate"]] & arguments[["help"]]){
...
+  end_year_month <- arguments[["<end_year_month>"]]
+  save_folder_path <- arguments[["<save_folder_path>"]]
 
-  mod_mbox_url <- conf[["mailing_list"]][["domain"]]
-  mailing_list <- conf[["mailing_list"]][["list_key"]][1]
-
-  start_year <- arguments[["<start_year>"]]
-  end_year <- arguments[["<end_year>"]]
+  conf <- yaml::read_yaml(conf_path)
+  mailing_list <- conf[["mailing_list"]][["mod_mbox"]][["project_key_1"]][["mailing_list"]]
```
carlosparadis, line 67: get on all of these hardcoded paths. it will cascade to the execs for the same reason they cascade to the notebooks.

carlosparadis: @RavenMarQ fyi for your task

exec/mailinglist.R
```diff
@@ -51,56 +47,62 @@ if(arguments[["tabulate"]] & arguments[["help"]]){
...
 
-  mbox <- download_mod_mbox(base_url = mod_mbox_url,
-                            mailing_list = mailing_list,
-                            from_year=start_year,
-                            to_year=end_year,
-                            save_file_path = save_path,
-                            verbose = TRUE)
+  download_mod_mbox(
```
carlosparadis, line 69: should use refresh. exec is for refresh only for downloaders that have them

exec/mailinglist.R
```diff
@@ -51,56 +47,62 @@ if(arguments[["tabulate"]] & arguments[["help"]]){
...
-                            from_year=start_year,
-                            to_year=end_year,
-                            save_folder_path = save_path,
-                            verbose = TRUE)
-
-  cli_alert_success(paste0("Downloaded mailing list was saved at: ",save_path))
+  conf <- yaml::read_yaml(conf_path)
+  mailing_list <- conf[["mailing_list"]][["pipermail"]][["project_key_1"]][["mailing_list"]]
+
+  download_pipermail(
```
carlosparadis, line 91: refresh missing here too. if the download is the refresh itself, need to separate the download of one file in one function that is being called from refresh (i believe, check with me on call. it should mimic how jira api does its thing as explained on its notebook for consistency and sanity)

tools.yml
daomcgill: The path that is merged on main is the same as this. What should it be changed to?

exec/mailinglist.R
beydlern: The corresponding getter functions for these two LOC (88 and 89) are:
`conf <- parse_config(conf_path)`
`mailing_list <- get_pipermail_domain(conf, "project_key_1")`

exec/mailinglist.R
beydlern: The corresponding getter functions for these two LOC (66 and 67) are:
`conf <- parse_config(conf_path)`
`mailing_list <- get_mbox_domain(conf, "project_key_1")`

daomcgill: Thank you @beydlern 
@carlosparadis I now have it using the getter functions, but the directory is not created if it does not exist, i.e. I have to manually create the folders for ../../rawdata/helix/mod_mbox/save_mbox_mail/ before I can call the download functions. I recall you telling me to not check for the existence of directories in my mail functions, as this would be done somewhere else in the configuration. Should I be doing something about this?

exec/mailinglist.R
daomcgill: Changed to refresh

R/mail.R
daomcgill: Replaced with message using example from jira.R

carlosparadis:
@daomcgill look! the check for CMD is passing!! What sorcery you did to fix this thing? I never seen this test error before. Hmm..

daomcgill:
@carlosparadis I wish I could tell you which change I made to make the CMD check pass, but I really have no clue. Hopefully I can get the tests working next. The error message is very strange.

tools.yml
carlosparadis: Oops... I guess I confused myself here. Please go ahead and revert to what main has. I actually think I may have broken the Actions doing this, sorry! See if this makes both checks pass!

carlosparadis:
The R version change is a trap: I see the error reverted back to the one of my first fix a month ago:

>   Package required but not available: ‘XML’

XML is not available before 4.4 anymore was my conclusion. At 4.5 other type of error appeared. Also, I'd revert back to the commit at least your unit test worked, now both are failing!

daomcgill:
@carlosparadis I changed it back to the version that was passing the check. Now it is not working with the same version that was passing the checks before.
Locally, I had the exact same errors in my tests, which resolved themselves when I changed both the tools in kaiaulu/ and the one in the tests to have working paths to utags and perceval. Not sure where to go from here.

carlosparadis:
At the rate this is going, we might as well call an exorcist. Focus on your M3, and I will take a look at this again. Don't commit again so we don't enter in a merge conflict.

daomcgill:
@carlosparadis Sounds good. I resolved most of the review points, but did not change the notebook narrative. Should I work on that before going back to M3?

carlosparadis:
I would say work on it later on. Let's wrap the functionality of M3 first and use that as a backup task!

carlosparadis:
Starting review this PR now, please hold on commits! : )

R/mail.R
```diff
@@ -4,238 +4,566 @@
...
+      # Make sure the month has two digits.
+      month_str <- sprintf("%02d", month)
+      # Create a string in the format "YYYY-MM"
+      year_month_str <- sprintf("%04d-%02d", year, month)
+      # This constructs the URL from which the mbox for the current year and month will be downloaded.
+      # The format for the URL is fixed by Apache's Pony Mail service.
+      download_url <- stringi::stri_c("https://lists.apache.org/api/mbox.lua?list=", mailing_list_name, "&date=", year_month_str)
+
+      # Create the file name where the mbox will be saved locally, in the format ''kaiaulu_'YYYYMM.mbox'.
+      file_name <- stringi::stri_c("kaiaulu_", year, month_str, ".mbox")
```
carlosparadis, line 425: prefix hardcoded to kaiaulu. @daomcgill who is code reviewing this and executed your notebook?

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,252 @@
...
+With the configurations loaded, we can proceed to download the mailing list archives. The downloaders are responsible for fetching the archives from the specified mailing lists and saving them locally in .mbox format.
+
+### Pipermail Downloader
+
+For Pipermail, we need to specify the project key, which is used to retrieve the configuration parameters for the specific project. The project key is used to identify the project in the configuration file.
+
+Now, we can use the getter functions to retrieve the configuration parameters for the specified project key.
+
+```{r eval=FALSE}
+conf <- parse_config("../conf/helix.yml")
```
carlosparadis, line 116: This is incorrect. The config pipermail points to openssl. You can't use openssl config as helix to run the notebook. Use openssl config instead.. @daomcgill This needs to be properly executed and inspected.

carlosparadis (changes requested):
There are some issues here. Wait until I am done with the other fixes and post a comment so you can take over, but this will need to be addressed for the notebook to be merged.

conf/helix.yml
```diff
@@ -48,26 +48,31 @@ version_control:
...
-#       # mbox_file_path is for use only with parse_mbox() function. It is the file to parse
-#       mbox_file_path: ../../rawdata/kaiaulu/pipermail/save_mbox_mail/kaiaulu.mbox
-#     project_key_2:
-#       mailing_list: https://mta.openssl.org/pipermail/openssl-project/
-#       save_folder_path: ../../rawdata/kaiaulu/pipermail/save_mbox_mail_2/
-#       # mbox_file_path is for use only with parse_mbox() function. It is the file to parse
-#       mbox_file_path: ../../rawdata/kaiaulu/pipermail/save_mbox_mail_2/kaiaulu.mbox
+      save_folder_path: ../../rawdata/helix/mod_mbox/save_mbox_mail
+      # mbox_file_path is for use only with parse_mbox() function. It is the file to parse
+      mbox_file_path: ../../rawdata/helix/mod_mbox/save_mbox_mail/kaiaulu_202407.mbox
```
carlosparadis, line 58: @daomcgill @beydlern @crepesAlot what is going on with this mbox_file_path?

carlosparadis:
https://github.com/sailuh/kaiaulu/blob/513a3f08514421160a0f58d5b7461fa0d870169e/R/config.R#L151-L173

@daomcgill @crepesAlot @beydlern there is no such a thing as "mbox input file". Where is this coming from?

Helix config says:

"# mbox_file_path is for use only with parse_mbox() function. It is the file to parse" 

However, this does not make sense. We download several .mbox files in a folder, so that parse_mbox() can parse all the files in said folder to a table. Why would we hardcode a single file to the config?

daomcgill:
@carlosparadis To my understanding, the parse_mbox function only parses one file at a time. It takes as input a file name and not a folder (this was the way it was from the start, I did not change). I also structured parallel jobs to work according to this.

daomcgill:
@carlosparadis parse_mbox will not work without mbox file path

carlosparadis:
@daomcgill That is correct, however, you should not hardcode one out of many files in the folder into the config. The correct way to go about this is to, in your notebook, to use list.files to then loop through them calling parse_mbox() on each of them into a list, and then using rbindlist. Your exec and parallel should be file only, or you can't parallelize so that is correct.

What that being said, the parse_mbox() of this branch does not work. It throws a JSON error. The one on master works on your downloader. I noticed that you made some changes on parse_mbox, like addin a flag for mbox_dir. I tried to edit that out but it is still not working for me.

carlosparadis:
To be clear: The exec/parallel also should only assume the folder containing the files on the folder, and then figure out from there how to access the files. No single filepath should be hardcoded to the config. 

@beydlern and I had a recent discussion about this on his end too.

R/mail.R
```diff
@@ -247,30 +575,57 @@ download_mod_mbox_per_month <- function(base_url, mailing_list, from_year, to_ye
...
-                             args = c('mbox',mbox_uri,mbox_path,'--json-line'),
-                             stdout = TRUE,
-                             stderr = FALSE)
-  # Parsed JSON output as a data.table.
-  perceval_parsed <- data.table(jsonlite::stream_in(textConnection(perceval_output),verbose=FALSE))
-
-  columns_of_interest <- c("data.Message.ID","data.In.Reply.To","data.Date","data.From","data.To","data.Cc","data.Subject","data.body.plain","data.body")
-  columns_rename <- c("reply_id","in_reply_to_id","reply_datetimetz","reply_from","reply_to","reply_cc","reply_subject","reply_body","reply_body")
+  mbox_file_path <- path.expand(mbox_file_path)
+  mbox_dir <- dirname(mbox_file_path)  # Extract directory path
```
carlosparadis, line 589: @daomcgill why was this mbox_dir added?

R/mail.R
```diff
@@ -247,30 +575,57 @@ download_mod_mbox_per_month <- function(base_url, mailing_list, from_year, to_ye
...
+
+
+  if (length(json_lines) == 0) {
+    stop("No valid JSON lines found in Perceval output. Check the mbox file or Perceval configuration.")
+  }
+
+
+  # Parse JSON output as a data.table
+  perceval_parsed <- tryCatch({
+    data.table(jsonlite::stream_in(textConnection(json_lines), verbose = FALSE))
```
carlosparadis, line 615: @daomcgill I just noticed you are filtering these { outof the json before calling the json parser. Why? This was not in the original.

R/mail.R
```diff
@@ -247,30 +575,57 @@ download_mod_mbox_per_month <- function(base_url, mailing_list, from_year, to_ye
...
+  mbox_file_path <- path.expand(mbox_file_path)
+  mbox_dir <- dirname(mbox_file_path)  # Extract directory path
+  mbox_uri <- mbox_file_path  # URI points to the mbox file
+
+  # Use Perceval to parse the mbox file
+  perceval_output <- tryCatch({
+    system2(perceval_path,
+            args = c('mbox', mbox_uri, mbox_dir, '--json-line'),
+            stdout = TRUE,
+            stderr = TRUE)
```
carlosparadis, line 597: this stderr = TRUE is why it no longer was working for me. I am changing it back to FALSE on the next commit.

carlosparadis:
@daomcgill i am still editing so don't commit anything, but this is what I meant: [d3dd232](https://github.com/sailuh/kaiaulu/pull/295/commits/d3dd2329f3ee138b65eea6a32b39cfb3bc32782f)

carlosparadis:
Note: Documentation and all notebooks of master compile on commit [d3dd232](https://github.com/sailuh/kaiaulu/pull/295/commits/d3dd2329f3ee138b65eea6a32b39cfb3bc32782f).

carlosparadis:
@daomcgill hi Dao,

I think this is mostly ready for merge, but there is one thing missing I will need to pass back to you: The execs require a start_date. This will basically kill its main purpose, which is to run server-side on a cron job. 

The logic of the refresh should make the start_date optional (which means the refresh function itself should make the parameter optional). If no start_date is specified, then an attempt to read an existing file to the folder should be made (since there is no other way to know the start date). IF that fails, then an error should issue saying "No existing data is located. Please specify a start date". Otherwise, the function should proceed as usual.

We need this because server-side the start date is not going to be specified manually every time: The first is specified with a date, but thereafter it is the server itself that will call the script every day to keep the data up to date. If I got confused and this is already implemented let me know.

Please feel free to commit to this PR, I will not make any other changes to avoid conflict.

You should only need to edit the refresh of both pipermaill and mod mbox (do not pass the start_date optional logic to the actual downloader, this is refresher behavior only). Avoid editing anything else to keep the last revision focused so I can merge this right after.

I made a couple comments on the commits if you would like to see what I modified and what I was hoping for in the end. Note I edited the file names too, they are no longer kaiaulu_*. The prefix was confusing since it makes look like every mbox is for Kaiaulu. Since there is no way to obtain the project name here without passing more parameters, only YYYYMM.mbox remains. 

Thank you for your hard work on this too. I am glad to finally have downloaders and refreshers for these.

carlosparadis:
@crepesAlot  @nicoelee123  please confirm if this file path was added to every config project_key. I am referring to `get_mbox_input_file` is there any equivalent on the other downloaders? If so, they should be removed. The folder path should be the one used instead. I will eventually remove the functions, so I need to know them.

Similar question for the project config file, are they specified anywhere else other htan openssl.yml and helix.yml?

daomcgill:
@carlosparadis 
So far, I have changed the refreshers to look for a start_year_month in the existing files, and return an error if there are no existing files and no start_year_month.

I have a question about what you said here:
``` "Your exec and parallel should be file only, or you can't parallelize so that is correct." ```
and here:
``` "To be clear: The exec/parallel also should only assume the folder containing the files on the folder, and then figure out from there how to access the files. No single filepath should be hardcoded to the config." ```

In the Python script for parallelization, I call the parse functions (parse exec) one time per file, since this is the easiest way to separate jobs. If I were to pass a folder to the parse exec, and handle the logic for parsing each file within there, obviously this would not work. How should I deal with this?

**Comment by daomcgill** [comment]
@carlosparadis Only solution I can think of is to have the exec check whether the input is file or folder. Would that work for you?

**Comment by carlosparadis** [comment]
@daomcgill Sorry for the confusion, what I meant to say was that I agreed with you: The parallel should be file only. It is already correct, I worded it wrongly. The parse function should be at file level too. The only thing I want to remove is the hardcoded file from the config, that was the part that did not make sense to me.

The notebook should list the files and parse. One thing I'd appreciate you checking is the behavior of parse_mbox when pointed to a folder: I have added code in one of the notebooks that properly look through each file and call parse_mbox() on each file. I have also left a few notebooks that call it just calling the function on the folder. In this latter case, when I inspected the table it seemed to me it did not parse all files. Could you check what it is doing?

**Comment by daomcgill** [comment]
@carlosparadis when I call parse_mbox from a loop in the notebook, all files appear to be parsed. I could not run the loop in the social smells notebook you edited, since I am missing data for it. Would you like me to add a loop example in the mail notebook?

**Comment by carlosparadis** [comment]
@daomcgill In the new notebook download_mail  I included the parse functions to showcase the downloader results through the parser. If you confirm pointing parse_mbox() to a folder lead to weird results not taking all files, I would recommend you add the loop not only to both parsers, but also to the other notebooks that use parse_mbox() in Kaiaulu, as they will be abiding by the new format.  See the loop I added on [d3dd232](https://github.com/sailuh/kaiaulu/pull/295/commits/d3dd2329f3ee138b65eea6a32b39cfb3bc32782f). 

For the notebooks, i'd say search parse_mbox() on GitHub Kaiaulu repo for hits. The notebooks that had _ are a good start (I changed in one of them already).

**Comment by daomcgill** [comment]
@carlosparadis Oh! Now I understand. I just tried passing a folder into parse_mbox() from the mail notebook, and it does actually parse all files. I checked the parsed rows against the number of unique reply IDs, and they were equal.

File: vignettes/download_mail.Rmd [hunk 38]
```diff
@@ -0,0 +1,215 @@
...
+)
+
+```
+
+After running this function, the .mbox files will be saved in the specified directory with filenames like 202310.mbox, 202311.mbox, etc, which can be parsed in a table:
+
+```{r}
+parsed_mail <- parse_mbox(
+  perceval_path = parse_perceval_path,
+  mbox_file_path = mbox_file_path
```
beydlern, line 134: I believe this is meant to parse the pipermail data which is stored in the pipermail folder path, however this is attempting to parse data from the mbox folder path: 
`mbox_file_path = mbox_file_path` should be changed to `mbox_file_path = pipermail_save_folder_path`.

vignettes/download_mail.Rmd
```diff
@@ -0,0 +1,215 @@
...
+
+```{r}
+# Load tools configuration
+tools <- parse_config("../tools.yml")
+parse_perceval_path <- get_tool_project("perceval", tools)
+
+# Load project configuration
+conf <- parse_config("../conf/helix.yml")
+mbox_file_path <- get_mbox_path(conf, "project_key_1")
+```
```
beydlern, line 91: Currently, folders need to be manually created. I believe this will be resolved later, but right now this remains an inconvenience for the user. The function to create these folders may be found in the PR #333.

beydlern (changes requested):
Barring the minor changes requested, the notebook is clear in its presentation and runs as expected.

R/mail.R
beydlern: My current issue #317 has prevented me from doing a thorough review of Dao's notebook. I just posted my review and requested a few changes.

carlosparadis:
@daomcgill could you resolve the merge conflicts?

daomcgill:
@carlosparadis resolved!

beydlern (approved):
Finished review.

carlosparadis:
@connorn-dev this does not look right. Are you sure you are sending the commits to the right Pull Request?

carlosparadis:
The exec of github.R you are extending is in https://github.com/sailuh/kaiaulu/pull/301. I am not sure why you have been sending commits to the mailing list downloader. Dao's code was only to be used as referral. 

Let me fix this PR once you commit the code to the appropriate one. Do not rewind this branch, I do not want to risk losing her code.

connorn-dev:
I apologize for the confusion, I completely overlooked the branch reference
in my issue. My issue has this PR to be updated but I will change
that. I have the code so feel free to fix the PR, I will update #301
<https://github.com/sailuh/kaiaulu/pull/301> with the CLI code. For the
vignette should I push to the same PR or create a new one?

Thank you

carlosparadis:
@connorn-dev extend #301. For now keep github.R separately, but we should discuss later if it makes sense combining them into one file. No worries, just ping me on #301 when you have all the code there and i will fix this one.

carlosparadis:
@geraldmjhuff I had to do a force push to erase the incorrect commits pushed here. If you are using this branch, you may need to git fetch it again. At the very least, the process mining code should no longer get in the way.

Issue #296, opened 2024-04-22
JIRA Refresh Script


Ssunoo2:
The purpose of this is to make a script for the JIRA refresh that will go in the exec file. This script should be runnable from the command line given certain input values. The ultimate goal of this script is to be run hourlly to continuously be updating issue JIRA data.

PR #297, opened 2024-04-23
i #296 Jira exec works from commandline

Ssunoo2:


exec/jira.R
```diff
@@ -0,0 +1,95 @@
...
+require(data.table,quietly=TRUE)
+
+
+# Rscript jira.R refresh issues ../../kaiaulu/tools.yml ../../kaiaulu/conf/geronimo.yml ../../rawdata/issue_tracker/geronimo/issue_comments/
+
+doc <- "
+USAGE:
+  jira.R refresh help
+  jira.R refresh issues <tools.yml> <project_conf.yml> <save_file_name_path>
+  jira.R download modmbox help
```
carlosparadis: Note to self: Should remove this from here now we have a mailinglist exec that handles this with refresher.

Issue #313, opened 2024-10-09
Expanding the Syntax Extractor


daomcgill:
-----

## Purpose 

The Syntax Extractor in Kaiaulu is used to extract meaningful information from source code using srcML. The purpose of this task is to extend the syntax extraction capabilities by adding new functions to extract file-level and class-level documentation using XPath queries. The extracted data will be used in future stages, where the goal is to combine the data with NLP to create semantic representations of the code.

## Process
1. Understand the current syntax extractor functions, and how annotations are represented in srcML XML files.
2. Understand XPath queries, and create queries for the new functions (file and class level documentation).
3. Implement new functions with the custom XPath queries.
4. Test on various examples.
5. Consider other syntactic elements that could be useful for the bigger picture. Example: Functions in git.R that retrieve commit messages or issue discussions tied to a file.
6. Create a notebook for Syntax Extraction, and maintain it with existing/ new functionality.

## Existing Functions
- [annotate_src_text()](https://github.com/sailuh/kaiaulu/blob/c7811069ef3c7943b509c76d4b9116d1a14dc343/R/src.R#L296): Runs srcML on a folder of source code and outputs the annotated XML.
- [query_src_text()](https://github.com/sailuh/kaiaulu/blob/c7811069ef3c7943b509c76d4b9116d1a14dc343/R/src.R#L327): Runs an XPath query on the annotated XML generated from annotate_src_text().
- [query_src_text_class_names()](https://github.com/sailuh/kaiaulu/blob/c7811069ef3c7943b509c76d4b9116d1a14dc343/R/src.R#L356): Extracts class names from the annotated source code.
- [query_src_text_namespace()](https://github.com/sailuh/kaiaulu/blob/c7811069ef3c7943b509c76d4b9116d1a14dc343/R/src.R#L399): Extracts the namespace (file path) from the annotated source code.

## New Functions
- query_src_text_file_documentation(): Extract file-level documentation (e.g. comments in the file header).
- query_src_text_class_documentation(): Extract class-level documentation (e.g. comments before class declaration).

## Task List 
- [ ] Install and set up [srcML](https://www.srcml.org/doc/java_srcML.html). This takes source code and add XML annotations (e.g. classes, methods, variables, etc.). Include the path to srcML in tools.yml.
- [ ] Use the existing functions to generate XML files. Inspect the files to understand the output.
- [ ] Gain an understanding of how XPath queries are formed, and figure out how to create queries for the new functions.
[...]

PR #320, opened 2024-10-15
i #313 Added Notebook for Syntax Extraction

daomcgill:
- Documented current syntax extraction functions
- Overview on syntax extraction and XPath
- Placeholder for syntax.yml config file

### DIFF
--- a/conf/syntax.yml
+++ b/conf/syntax.yml
@@ -1,0 +1,71 @@
+# -*- yaml -*-
+# https://github.com/sailuh/kaiaulu
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Project Configuration File #
+#
+# To perform analysis on open source projects, you need to manually
+# collect some information from the project's website. As there is
+# no standardized website format, this file serves to distill
+# important data source information so it can be reused by others
+# and understood by Kaiaulu.
+#
+# Please check https://github.com/sailuh/kaiaulu/tree/master/conf to
+# see if a project configuration file already exists. Otherwise, we
+# would appreciate if you share your curated file with us by sending a
+# Pull Request: https://github.com/sailuh/kaiaulu/pulls
+#
+# Note, you do NOT need to specify this entire file to conduct analysis.
+# Each R Notebook uses a different portion of this file. To know what
+# information is used, see the project configuration file section at
+# the start of each R Notebook.
+#
+# Please comment unused parameters instead of deleting them for clarity.
+# If you have questions, please open a discussion:
+# https://github.com/sailuh/kaiaulu/discussions
+
+project:
+  website: https://github.com/junit-team/junit5/
+  #openhub: https://www.openhub.net/p/apache_portable_runtime
+
+version_control:
+  # Where is the git log located locally?
+  # This is the path to the .git of the project repository you are analyzing.
+  # The .git is hidden, so you can see it using `ls -a`
+  log: ../rawdata/git_repo/junit5/.git
+  # From where the git log was downloaded?
+  log_url: https://github.com/junit-team/junit5/
+  # List of branches used for analysis
+  branch:
+    - main
+
+filter:
+  keep_filepaths_ending_with:
+    - cpp
+    - c
+    - h
+    - java
+    - js
+    - py
+    - cc
+  remove_filepaths_containing:
+    - test
+    - java_code_examples
+
+tool:
+ # srcML allow to parse src code as text (e.g. identifiers)
+  srcml:
+    # The file path to where you wish to store the srcml output of the project
+    srcml_path: ../analysis/depends/srcml_depends.xml
+# Analysis Configuration #
+analysis:
+  # A list of topic and keywords (see src_text_showcase.Rmd).
+  topics:
+    topic_1:
+      - model
+      - view
+      - controller
--- a/vignettes/syntax_extractor.Rmd
+++ b/vignettes/syntax_extractor.Rmd
@@ -1,0 +1,305 @@
+---
+title: "Syntax Extractor"
+output: 
+  html_document:
+    toc: true
+    number_sections: true
+vignette: >
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteIndexEntry{Kaiaulu Syntax Extractor}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r eval=FALSE}
+rm(list = ls())
+seed <- 1
+set.seed(seed)
+
+# Load libraries
+  require(kaiaulu)
+  require(data.table)
+  require(yaml)
+  require(stringi)
+  require(XML)
+  require(gt)
+```
+
+# Introduction
+
+In open-source projects, code is often spread across many files and organized in various ways. Understanding this code at a glance can be difficult, especially when projects grow large and complex. This is where syntax extraction comes in.
+
+The Syntax Extractor is a tool that helps us dive into the structure of source code by using an annotated format. It works by converting code into a structured XML format using srcML, making it easier to analyze and understand the underlying architecture of the project. By extracting key pieces of information such as class names, functions, or even comments (like documentation), we can generate meaningful data that helps with things like semantic analysis, code comprehension, and more.
+
+Imagine a codebase as a collection of books in a library. Without a catalog, finding the right book would be nearly impossible. Syntax extraction is like building a catalog for a codebase. It allows us to quickly pinpoint where classes, methods, and important comments are located, and organize this data in a structured way.
+
+For Kaiaulu, this is particularly useful as we move toward deeper analysis, such as applying machine learning techniques like word embeddings to the extracted syntax. This allows us to not only understand how the code is written but also how the pieces relate to each other on a semantic level.
+
+In short, syntax extraction gives us a way to “see the big picture” without having to manually dig through every line of code.
+
+## How the Syntax Extractor Works
+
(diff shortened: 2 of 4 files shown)


carlosparadis (changes requested):
These should be double `../../` because when we run the code block, it assumes it is located on vignettes. I believe this also needs to be fixed on the mail PR.

conf/syntax.yml
carlosparadis, line 1: This conf name does not make sense, the config files are per project. Syntax is not a project!

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+  require(data.table)
+  require(yaml)
+  require(stringi)
+  require(XML)
+  require(gt)
+```
+
+# Introduction
+
+In open-source projects, code is often spread across many files and organized in various ways. Understanding this code at a glance can be difficult, especially when projects grow large and complex. This is where syntax extraction comes in.
```
carlosparadis, line 29: I would remove this. The motivation fits better to why someone would need Kaiaulu than the feature you are working here!

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+The Syntax Extractor is a tool that helps us dive into the structure of source code by using an annotated format. It works by converting code into a structured XML format using srcML, making it easier to analyze and understand the underlying architecture of the project. By extracting key pieces of information such as class names, functions, or even comments (like documentation), we can generate meaningful data that helps with things like semantic analysis, code comprehension, and more.
```
carlosparadis, line 31: You should reword this to not give the allusion this Notebook is a tool of its own. For example, I would expect this to be on a README of a standalone tool with the feature you are working on, rather than the explanation of a new capability in Kaiaulu.  

Rather you could start by something like: `This notebook explains how to convert` and then tie to what you have on `code into a structured XML format using the tool srcML`. 

I would also remove the `making it easier to analyze and understand the underlying architecture of the project`. Rather, the motivation is to `selectively extract terms of a source file, so it may be represented for other analysis, such as`. and connect to `semantic analysis, code comprehension, etc.`

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+Imagine a codebase as a collection of books in a library. Without a catalog, finding the right book would be nearly impossible. Syntax extraction is like building a catalog for a codebase. It allows us to quickly pinpoint where classes, methods, and important comments are located, and organize this data in a structured way.
```
carlosparadis, line 33: Analogies are very good, but I'm not sure if we want to "sell the fish" here as syntax extraction being a way to catalogue source code to find classes. Instead, I'd like you to try another analogy, that builds on your original motivation of semantic analysis and code comprehension. 

Food for thought for the analogy: Why would we want to do semantic analysis and code comprehension assessments from extracting the file?

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+For Kaiaulu, this is particularly useful as we move toward deeper analysis, such as applying machine learning techniques like word embeddings to the extracted syntax. This allows us to not only understand how the code is written but also how the pieces relate to each other on a semantic level.
```
carlosparadis, line 35: This comes closer to the spirit of the analogy: Why would we want a machine to understand semantically how two classes relate? How would this be helpful for a developer?

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+In short, syntax extraction gives us a way to “see the big picture” without having to manually dig through every line of code.
```
carlosparadis, line 37: I would like you to emphasize a bit more the fact that, representing a file through different methods, can lead to different representations, and consequently understanding by a computer that is comparing files using this information!

For example, consider that we could a) show all the file content to a computer or b) only the names of classes, variables and documentation, which capture semantic meaning tied to the code, or c) commits and issues discussion associated to the file (just like we are doing now), which capture others discussion surrounding the file. 

You could give a computer different representations (a, b, c) of a given file, and ask the computer to compare to another file (also represented in either a, b, c). Consider a computer would be having a different understanding of what two files are related depending on the representation of your choice:

It may say using criteria c) that the files are alike, because generally people are discussing similar features on its commits and issues, but if using a) the way the files were implemented being substantially different, they could be different.

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+## How the Syntax Extractor Works
+
+At its core, the syntax extractor relies on [srcML](https://www.srcml.org/), which is a tool that converts source code into an XML representation. This XML structure gives us a detailed breakdown of the code: classes, functions, variables, and comments are annotated with specific XML tags, allowing us to query and extract what we need.
```
carlosparadis, line 41: I would defer this to after the project configuration file, so you can explain how it works as you call functions bit by bit, interspersed with examples and runnable code.

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+In this notebook, we’ll walk through the process of setting up the syntax extractor, running the extraction process, and then querying the annotated code for useful information like class names, namespaces, and documentation.
+
+# Project Configuration File
+
+In a project, source code is spread across multiple repositories and folders. To use Kaiaulu’s syntax extraction functions, you need to configure the system with details about where the source code is stored and where the tools (like srcML) are located.
+
+Kaiaulu uses a project configuration file format to specify the paths and settings necessary for syntax extraction. This allows you to manage different projects and their associated codebases efficiently.
```
carlosparadis, line 49: I would replace this with more precise motivation. For example, `manage different projects codebases efficiently` does not tell me much. 

For those two sentences, you want to focus on the fact that a) Kaiaulu project configuration files contain information about a **project** that is of interest, b) where you want to store the input and output to the tool, and c) what configuration parameters you want out of the tool (i.e. the tool may have its own arguments, the config is where you store them). The reason is reproducibility. If I don't know what you used as input, and the arguments, then I can't re-run and expect the same result.

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+tool:
+  srcml:
+    srcml_path: /usr/local/bin/srcml
+    srcml_output_path: ../../analysis/maven/srcml_output.xml
+
+analysis:
+  topics:
+    - class
+    - method
+    - documentation
```
carlosparadis, line 74: this probably will need revisiting later, it is legacy of another thing we were doing with srcML. We can go over this on call

vignettes/text_gof_showcase.Rmd
carlosparadis, line 1: should undo the `../` changes

vignettes/syntax_extractor.Rmd
```diff
@@ -0,0 +1,305 @@
...
+)
+
+# Display the extracted comments
+function_comments %>%
+  gt()
+```
+
+This query will return a list of comments associated with each function in the code.
+
+### XPath Cheat Sheet
```
carlosparadis, line 221: We should make an actual Cheatsheet for Kaiaulu for this later on.

carlosparadis (changes requested):
I have not done the full review yet, but I hope this helps. Is my understanding correct that the documentation extractor is still pending? 

We can do another full pass on the notebook at that point after you try it out.

conf/syntax.yml
daomcgill: Will change the name. I am a little bit confused on the ../../ since I thought we were supposed to set out Global Options -> R Markdown -> Evaluate chunks in directory -> Project?

daomcgill:
@carlosparadis I made the changes we discussed. I also added functions for extracting function names (opt. params) and variable names (opt. types). I do need to do a more thorough manual check to make sure the results are accurate, but it appears to be working so far. 
Should I continue adding functions--maybe imports or some way of determining inheritance structure (I would need to think about this some more)--or move on and come back once the other pieces start to come together?

carlosparadis:
Working on this PR now, please hold on doing commits.

carlosparadis:
@daomcgill i suggest you fix this PR first with the get and the other notes, assuming this is the one that came before the exec PR. Then you can merge locally this branch to the exec PR, to avoid having to do it twice (just be careful you dont merge the one that came after to the PR before or it will mix everything up).

daomcgill:
@carlosparadis okay, I will work on this now

daomcgill:
@carlosparadis fixed hardcoded paths and reverted workflow. Going to move to 314 now

carlosparadis:
@daomcgill could you confirm the notebook display the parsed tables so the users can see what each function outputs as i did with parse_mbox on your mail notebook?

daomcgill:
@carlosparadis every query that I run displays a table. Should I allow every query to be evaluated then? I think that also means I have to change it so the annotation is evaluated?

daomcgill:
@carlosparadis I think I answered my own question. Notebook now displays tables.

carlosparadis:
@daomcgill you sure did! It is very much in line with the changes I did on mail and that you see on itm0, there is no better way to showcase your work on the docs than having a table showing on the intermediate steps. Please send me on drive the files needed to run the code (I imagine just the raw data suffice, the rest I can extract from your functions). Thank you!

daomcgill:
@carlosparadis I just used the cloned maven repository but, yes, I am adding it to the drive right now (folder is just named maven). 

I think I have made it through all the requested changes, in both branches. I have also updated the fasttext notebook, and it appears to be running with the new config setup.

carlosparadis:
Thank you!!

PR #325, opened 2024-10-27
317 Openhub API Interface Functions

beydlern:
- Created undocumented Ohloh API interfacing functions to access OpenHub's database.
- Rough openhub_project_search.Rmd notebook that allows the user to search for projects based on code language and organization name (`temp_max_number_of_portfolio_projects_pages_to_search_through` is a temporary variable (renamed in the future) that will limit the number of pages the `openhub_api_portfolio_projects` function will request to not consume too many API requests when testing, currently set to `1` (one page)).


daomcgill:
@beydlern I tried running through your notebook, and everything seems to be in working order. I think it looks good.

R/config.R
```diff
@@ -5,6 +5,495 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 
+########## OpenHub Functions / Ohloh API Interfacing Functions ##########
+
+#' Download a single page of Organization XML Response File.
+#'
+#' @description Downloads an XML response file from "organization" collection
```
carlosparadis, line 12: remove description tags, rest of codebase do not use them

R/config.R
```diff
@@ -5,6 +5,495 @@
...
+    warning(paste0("openhub_api_analyses: No current analysis collection found for project name: ", project_name))
+  } else {
+    warning(paste0("openhub_api_analyses: ", status)) # prints the status warning message
+  }
+
+  if (!is.null(timestamp)) {
+    openhub_download(api_response, save_folder_path, timestamp, type, unique_information=project_name)
+  }
+
+  invisible(api_response)
```
carlosparadis, line 198: what is the purpose of this function? first time seeing it being used

_pkgdown.yml
```diff
@@ -356,3 +356,19 @@ reference:
...
+- title: __OpenHub__
+  desc: >
+   Functions to interact and download data from OpenHub's API, Ohloh.
+   See [OpenHub Project Search](../articles/openhub_project_search.html)
+   Notebook for an example.
+- contents:
+  - openhub_api_organizations
+  - openhub_api_portfolio_projects
+  - openhub_api_projects
+  - openhub_api_analyses
```
carlosparadis, line 368: can you confirm all the api_ functions save the raw, unmodified file to disk? I am assuming they are XMLs, or was it JSONs? It is important the downloaders do not do any pre-processing to the data.

vignettes/openhub_project_search.Rmd
carlosparadis, line 1: Not all evals should be false. If you look on itm0, we show results here and there so users can make sense. You should use the downloader to download some files locally, specified in the config, and also call the parsers for each endpoint to display with gt. Please put on drive the files so I can download them too to run so I can test the docs re-compiled. Please give me an url over email to it.

carlosparadis: @daomcgill  @RavenMarQ @crepesAlot very important. I noticed dao notebook's also were missing the parser. For someone to look at the docs on itm0, they have to see a sample of the tables. It greatly help understand what the notebook can give them since without code, the notebook is basically english, tables and visualization.

vignettes/openhub_project_search.Rmd
```diff
@@ -0,0 +1,230 @@
...
+With the organization API response (only one page), we may parse this response with its corresponding parser function, `openhub_parse_organizations`, to acquire a data table with columns representing the tags for the organization listed:
+
+* name: The name of the organization.
+* html_url_projects: The URL to the organization on OpenHub's website corresponding to a list of portfolio projects for the organization.
+* portfolio_projects: The number of portfolio projects under the organization.
+
+```{r, eval = FALSE}
+openhub_organization_api_requests <- openhub_retrieve(organization_folder_path)
+openhub_organizations <- openhub_parse_organizations(openhub_organization_api_requests)
+gt(openhub_organizations)
```
carlosparadis, line 114: look at the gt that I did on the merged notebooks: this is missing the pipe with the head operator. Failing that this will be a massive table. @RavenMarQ @crepesAlot  @daomcgill please check your notebooks have that.

carlosparadis (changes requested):
Doing a quick review since I won't have any more time to test today. Note the comments are just examples, please check the entire PR for other ocurrences like description and the gt function call.

R/config.R
beydlern: The `invisible()` function just like a `return()`, but it doesn't print the output of the function if a call to the function isn't assigned to a variable. This will be removed as there is no need for the API functions to return the API response files as they're downloaded.

_pkgdown.yml
beydlern: The response files are all XML files and all of the openhub_api_* functions are downloaders that do not do any pre-processing to the data.

vignettes/openhub_project_search.Rmd
beydlern: Will be added.

PR #326, opened 2024-10-29
318 - Proposed Changes for GoF Motif Notebook

RavenMarQ:
Proposed changes and comments for the graph_gof_showcase notebook.

### DIFF
--- a/vignettes/graph_gof_showcase.Rmd
+++ b/vignettes/graph_gof_showcase.Rmd
@@ -16,19 +16,37 @@ This notebook provides a brief introduction to the early version of the **GoF** 
 
 # Setup
 
-## Graph Approach
-
-The graph approach uses Tsantalis [pattern4.jar](https://users.encs.concordia.ca/~nikolaos/pattern_detection.html). As with other third party tools, you should specify the path to the jar in the `tools.yml` and provide the necessary parameters of the project configuration file. 
-
 ```{r}
 rm(list = ls())
 seed <- 1
 set.seed(seed)
 ```
 
-# Project Configuration File
+
+```{r warning=FALSE,message=FALSE}
+require(kaiaulu)
+require(data.table)
+require(yaml)
+require(stringi)
+require(knitr)
+require(reticulate)
+require(magrittr)
+require(gt)
+```
+
+# The Graph Approach
+
+The graph approach uses Tsantalis [pattern4.jar](https://users.encs.concordia.ca/~nikolaos/pattern_detection.html). As with other third party tools, you should specify the path to the jar in the `tools.yml` and provide the necessary parameters of the project configuration file. 
+
+// PROPOSED EDIT: Leave The Graph Approach section as an explanation of the approach rather than delving into configuring the tools.yml. If it only serves as to point to where they should be getting the jar, maybe a reworking of this section would be necessary- while configuring tools.yml to point to their installed jar should be included in the next section. 
+
+# Project Configuration Files
+
+// PROPOSED EDIT: Inserting where and what I should be changing to point to the jar here. Something like: "Now with the pattern4.jar installed, in the Kaiaulu folder is a tools.yml file. Changing the pattern4 variable there..." would be sufficient in guiding the user. Then an explanation on what, how, why, etc. can be inserted afterwards.
 
 Analyzing open source projects often requires some manual work on your part to find where the open source project hosts its codebase and mailing list. Instead of hard-coding this on Notebooks, we keep this information in a project configuration file. Here's the minimal information this Notebook requires in a project configuration file:
+
+// PROPOSED EDIT: This part was confusing to me stepping in: I was confused whether I should find the configuration file (if it existed) or if I should edit the values in this yaml bit. A user should not have to read the code block (which requires an understanding of R) to glean where they should go (in this case tools.yml and /conf/junit5.yml respectively). Clearing up where they should be changing the files explicitly helps reduce the confusion, as well as explaining why we've put this yaml block below or possibly removing it from this notebook and having more explanatory comments in the junit5.yml.
 
 ```yaml
 project:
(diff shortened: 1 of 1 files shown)


vignettes/graph_gof_showcase.Rmd
```diff
@@ -16,20 +16,38 @@ This notebook provides a brief introduction to the early version of the **GoF**
```
daomcgill, line 16: The introduction section is very brief and could still leave the user questioning what the purpose is. Clarify that this notebook guides users through analyzing GoF patterns in external Java projects. It would probably be beneficial to explain GoF/Design Patterns and what they are/why we use them? Why would we want to analyze these patterns and how would the information be beneficial to the user?

vignettes/graph_gof_showcase.Rmd
```diff
@@ -16,20 +16,38 @@ This notebook provides a brief introduction to the early version of the **GoF**
...
+require(gt)
+```
+
+# The Graph Approach
+
+The graph approach uses Tsantalis [pattern4.jar](https://users.encs.concordia.ca/~nikolaos/pattern_detection.html). As with other third party tools, you should specify the path to the jar in the `tools.yml` and provide the necessary parameters of the project configuration file. 
+
+// PROPOSED EDIT: Leave The Graph Approach section as an explanation of the approach rather than delving into configuring the tools.yml. If it only serves as to point to where they should be getting the jar, maybe a reworking of this section would be necessary- while configuring tools.yml to point to their installed jar should be included in the next section. 
+
+# Project Configuration Files
```
daomcgill, line 43: I think the setup could be broken into two separate sections:
1. Preparing the Java Project: Java installation and SDK config, getting the project, step-by-step compilation. How can the user determine whether the project uses Maven or Gradle? If that changes compilation steps, how so? You mention you used intelliJ, but the user may need help with doing this as well. If you choose to walk them through compiling using intelliJ, I would include screenshots and links. I would also, at least briefly, mention command line compilation options since everyone has command line. Since this is meant to teach the user how to use the tool, consider adding links to one or two pre-compiled projects. You could then use these projects when walksing through examples. This will also help the user to know whether any problems that arise are in the compilation or somewhere else.
2. Configuration: what you mentioned below about configuration, also how the user should find the paths they need

vignettes/graph_gof_showcase.Rmd
```diff
@@ -76,17 +94,6 @@ tool:
...
-require(kaiaulu)
-require(data.table)
-require(yaml)
-require(stringi)
-require(knitr)
-require(reticulate)
-require(magrittr)
-require(gt)
-```
-
```
daomcgill, line 97: - After the configuration section, the notebook could be more explicit in stating what each command does. The output table is commented on, but does not really explain what the user is looking at. What does each row represent?
- The section on mapping to file paths could be separated, making clear that it is optional for users who want to see exactly where the patterns appear in the project. Should probably include what srcML is and how to install it.
- When you tried using this did it work for all projects, or only some? Testing it out on various projects should help with letting the user know how they can expect to use the tool. Be mindful that a process that takes you a long time could be near impossible for a user to navigate. The notebook should take the complex process, break it down, and make it easily digestible. The reason you should test on different projects is so the user does not have to. You can just tell them what works and what does not.

daomcgill (commented):
I left comments according to my understanding. Hope it helps.

vignettes/graph_gof_showcase.Rmd
carlosparadis: Suggestion here is to take a quick look on apache java projects (ask @beydlern ) if they have any guidance on how to compile. IntelliJ may be the hard ad-hoc route. E.g. here: https://github.com/spring-projects/spring-framework?tab=readme-ov-file#build-from-source 

I took a quick look and says Gradle is used and to type ./gradlew build. Gradle website, in turn, says it is a open source tool to compile java: https://gradle.org/. 


Apache Tomcat is another Java project. A quick inspection on the README leads me to this: https://github.com/apache/tomcat/blob/main/BUILDING.txt . Now the project tells me rather than Gradle, it needs Apache Ant. Quick look on Google for Apache Ant is that it is also a tool to build Apache Java projects: https://ant.apache.org/ 

Pick a few more, and you will have a decent idea on what is out there in the wild to compile java projects (again we are more interested in Apache projects). 

The README at the very least should brief the user that there are a couple different ways to compile them and provide some pointers. If most projects use Gradle, then we could check if the instructions for the projects are similar or all over the place. 

In R all you have to do is type "install.packages()", maybe gradle projects are easy enough we could recommend users of this tool to use it. Brute forcing through the setup will not help the user on the decision making process nor decrease the learning curve.

vignettes/graph_gof_showcase.Rmd
```diff
@@ -12,10 +12,35 @@ vignette: >
 
 # Introduction
 
-This notebook provides a brief introduction to the early version of the **GoF** module (in particular Tsantalis pattern4.jar), using some parts of the **Text** module to identify some of the Gang of Four (GoF) Design Patterns in source code [pattern4.jar](https://users.encs.concordia.ca/~nikolaos/pattern_detection.html).
+> "The Gang of Four (GoF) Design Patterns, introduced in the book “Design Patterns: Elements of Reusable Object-Oriented Software,” authored by Erich Gamma, Richard Helm, Ralph Johnson, and John Vlissides, provide a catalog of proven solutions to common design problems in software development. The GoF Design Patterns encourage best practices, code reusability, and the separation of concerns, aiding in the development of robust and scalable applications."
```
daomcgill, line 15: see comments posted in line for the PR Discussions

daomcgill (commented):
see comments posted in line for the PR Discussions

daomcgill (approved):
see comments posted in line for the PR Discussions

PR #375, opened 2026-03-09
Transform and Graph Standardization

### LINKED ISSUE #361: Define Consistent Transform Interfaces and Graph Abstractions Across Kaiaulu Pipelines
# Purpose

This project standardizes how Kaiaulu's data transform pipelines produce network outputs. Kaiaulu supports multiple data sources including Git, Jira, GitHub, mailing lists, and source code analysis tools. Before this work, the transform functions across these pipelines had evolved independently. They did not produce comparable outputs, they used inconsistent internal structures, and downstream analysis code often had to work around these differences at the notebook level.

The work done here defines a shared pipeline contract across all data sources:

```
parse_*
    ↓
transform_*_to_network
    ↓
model_(directed|bipartite|heterogeneous)_graph
    ↓
projections / exports / metrics
```

Every transform function now produces a typed graph object by building an explicit nodes table and edgelist table and passing them to the appropriate `model_*` constructor. This means downstream functions like projections, DSM export, and analysis metrics can operate on any graph object without needing to know which pipeline produced it.

# Process

## Git Log, Depends, Understand
As I began Milestone 1, I wanted to confirm my understanding of the code layout referenced in prior discussions. While `graph.R` exists in the `R` folder, `network.R` does not appear to exist in the current repository. Based on inspection of the codebase, it appears that functionality previously associated with `network.R` has been split across source-specific files such as `git.R`, `jira.R`, `github.R`, `mail.R`, and parts of `src.R`. In particular, the transform functions for the Depends and Understand pipelines appear to live in `src.R`.

To understand current behavior, I reviewed existing code, vignettes, and performed hands-on testing of the `Git`, `Depends`, and `Understand` pipelines. Based on this work, I created a diagram representing how each pipeline currently operates. The diagram shows where external CLI tools are used, how data is transformed in R, and where each pipeline terminates. The diagram reflects observed behavior rather than intended or future architecture.

GitDependsUnderstand_Transform_Pipelines drawio

Through this review, several cross-pipeline inconsistencies were identified. The `Git` pipeline constructs graph objects using `model_directed_graph()` and then applies projections and weighting, while the `Depends` and `Understand` pipelines stop at node and edge list outputs and do not construct graph objects inside Kaiaulu. Because of this, outputs from different pipelines are not directly interchangeable.

During testing of the `Depends` pipeline, schema and naming issues were encountered. In one case, `parse_dependencies()` produced a nodes table with a name column, while `transform_dependencies_to_network()` expected a filepath column, causing the transform to fail unless the column was manually renamed. A separate issue was observed where the Depends CLI produced a JSON file with a `-file.json` suffix instead of the expected filename, leading Kaiaulu to fail when parsing the output despite successful CLI execution.

Differences in how edge weights are handled were also observed. The weight_types parameter has different meanings across pipelines. In the `Depends` pipeline, it determines which dependency-type columns are summed to produce a numeric edge weight. In the `Understand` pipeline, it only filters dependency kinds without computing numeric weights. In the `Git` pipeline, weighting is applied later, after graph construction. As a result, edge weights are not directly comparable across data sources.

Finally, while reviewing and testing the `Git` entity-based temporal transforms, I observed that column renaming behavior differs depending on the mode being used, which makes transform behavior harder to reason about at the API level.

## GitHub, Jira, Mail

As a continuation of M1, I reviewed the `GitHub`, `Jira`, and `Mail` transform pipelines to understand how data moves from raw artifacts into structured outputs in the current Kaiaulu codebase. This review follows the same process I used earlier for the `Git`, `Depends`, and `Understand` pipelines. The goal was to document how the pipelines actually behave today, identify where they stop, and note any issues or inconsistencies I encountered while tracing the code. I focused on the boundaries of each pipeline rather than internal implementation details.

GitHubJiraMail_Transform_Pipelines drawio

For the `GitHub` pipeline, I examined the download functions that retrieve issues, pull requests, comments, and commits from the GitHub API. These functions operate independently and write raw JSON files to disk. All parsing and transformation logic is handled by the function `parse_github_replies`, which reads the downloaded JSON artifacts and normalizes them into a single replies table. This function combines parsing and transformation into one step and relies on several helper functions internally. There are no intermediate tables exposed, and the pipeline stops once the replies table is produced. One issue I noted is that the reply schema is assumed implicitly across helper functions, and there is no clear separation between parsing raw data and applying semantic transformations.

For the Jira pipeline, the structure is more clearly separated. Raw issue data is downloaded using `download_jira_issues`, after which `parse_jira` converts the JSON files into issues and comments tables. A second function, `parse_jira_replies`, then transforms those tables into a normalized replies table. The pipeline stops at that point. Compared to `GitHub` and `Mail`, `Jira` is the only pipeline that exposes intermediate tables. While this makes the transformation steps easier to follow, the replies transformation assumes that issues and comments are complete and correctly linked, and there is no explicit validation of those assumptions in the code.

The `Mail` pipeline follows a different pattern. Mailing list archives are downloaded as mbox files and then processed by the function `parse_mbox`. This function calls Perceval to extract message data and immediately normalizes it into a replies table. Parsing and transformation are fully combined, and the pipeline ends after the replies table is created. I noticed that much of the important logic, such as reply construction and identity handling, is implemented in helper functions that are not visible at the pipeline level. The pipeline also depends heavily on Perceval, but error handling and failure cases are not clearly surfaced.

Across all three pipelines, the final output is a replies table, and none of the pipelines perform graph construction or projection. Even though the replies table is compatible with later graph modeling steps, those steps are not part of these pipelines as implemented. Any use of graph utilities happens outside of these pipelines. Another issue I observed is that refresh functions exist for extraction in each pipeline, but they operate independently and are not integrated into the main transformation flow, which makes them easy to miss when reading the pipeline logic.

Overall, the `GitHub`, `Jira`, and `Mail` pipelines function as ingestion and normalization pipelines. They successfully standardize different data sources into a common replies format, but they do so with inconsistent structure and implicit assumptions. Documenting these behaviors and boundaries is important so that downstream workflows do not assume functionality that is not actually present in the current implementation.

## Experiment: Applying `model_directed_graph()` to Depends

As part of Milestone 1, I ran an experiment applying `model_directed_graph()` to the output of the Depends pipeline. The objective was to determine whether Depends could be elevated to the same abstraction layer as GitLog and to identify what structural changes would be required.

The Depends pipeline currently proceeds as follows:

1. `parse_dependencies()` invokes the Depends CLI and parses its JSON output into a structured object containing file-level nodes and a dependency-type edgelist.
2. `transform_dependencies_to_network()` collapses the dependency-type matrix into a weighted file-to-file network, producing `nodes(name, color)` and `edgelist(from, to, weight)`.

At this point, the pipeline terminates and returns a weighted network representation.

In the experiment, I applied `model_directed_graph()` directly to the weighted edgelist returned by `transform_dependencies_to_network()`. This produced a `directed_graph` object containing:

- `nodes(name, type, color)`
- `edgelist(from, to, weight)`

Structurally, this matches the abstraction level produced by the GitLog pipeline.

Kaiaulu_GitLog_vs_Depends drawio

### Comparison: GitLog vs Depends

Comparing the revised Depends pipeline to GitLog reveals an important architectural difference. In the GitLog pipeline, graph abstraction is constructed inside the transform functions themselves. For example, `transform_gitlog_to_temporal_network()` generates an edgelist and immediately calls `model_directed_graph()`, after which projections and weighting logic are applied to the resulting graph object. The Git pipeline therefore transitions into the graph abstraction layer during transformation.

In contrast, the current Depends pipeline produces a weighted network list but does not enter the graph abstraction layer unless `model_directed_graph()` is applied explicitly afterward. The experiment demonstrates that Depends can cleanly align with GitLog once this additional step is introduced.

There is no structural difference in the inputs or outputs required by `model_directed_graph()` between the two pipelines. In both cases, the function expects an edgelist with `from`, `to`, and optionally `weight`. However, the semantic meaning differs. In GitLog, edges often represent developer-to-file relationships and may involve bipartite or temporal structures. In Depends, edges represent static file-to-file dependencies. The abstraction layer is therefore consistent, even though the domain semantics differ.

### Constructing a `model_undirected_graph()` Abstraction

During this comparison, it also became clear that Kaiaulu currently assumes directed graphs as the default abstraction. While this is appropriate for many pipelines (such as commit or dependency flows), some data sources are conceptually symmetric. For example, co-change networks, co-authorship networks, or certain projections derived from GitHub, Jira, or mail replies may be better represented as undirected graphs.

A `model_undirected_graph()` function could be constructed analogously to `model_directed_graph()`, but would normalize edge directionality before building the graph abstraction. Conceptually, this would involve:

1. Taking an edgelist with `from`, `to`, and optional `weight`.
2. Sorting or canonicalizing node pairs so that `(A, B)` and `(B, A)` are treated as the same edge.
3. Aggregating duplicate edges by summing weights where appropriate.
4. Constructing a graph object flagged as undirected.

The resulting abstraction would maintain the same standardized schema for nodes and edges but remove directionality from the representation. This would allow pipelines that naturally produce symmetric relationships to converge to the same graph abstraction contract without artificially imposing direction.

Introducing `model_undirected_graph()` would further support Milestone 2's goal of defining consistent transform interfaces. Instead of each pipeline implicitly deciding whether it is directed or undirected, the abstraction choice would be explicit and standardized at the graph-construction layer.

### Implications for Standardization

The experiment shows that the Depends pipeline can be aligned with GitLog at the graph abstraction level. Once `model_directed_graph()` is applied, both pipelines produce structurally consistent `directed_graph` objects with compatible node and edge schemas. The remaining differences are semantic rather than structural.

Introducing both `model_directed_graph()` and a potential `model_undirected_graph()` abstraction would allow all pipelines to converge on a shared contract:

```
parse_*  
    ↓  
transform_*_to_network  
    ↓  
model_(directed|undirected)_graph  
    ↓  
projections / exports / metrics  
```

Downstream projections, exports such as DSM, and analysis workflows could then operate uniformly on graph objects rather than pipeline-specific list structures.

This experiment, therefore, not only validates the feasibility of standardizing Depends but also suggests a broader architectural improvement for Kaiaulu's graph modeling layer moving forward.

## Transform Function Standardization

After the assessment phase, I refactored the `transform_*` functions across Kaiaulu to follow a consistent pattern. Each function now builds an explicit nodes table and an explicit edgelist table and passes them to the appropriate `model_*` constructor before returning. This replaces earlier implementations where some functions returned plain lists with no class, used old constructor parameters like `is_bipartite` and `color` that are no longer part of the API, or skipped graph construction entirely and left that responsibility to the caller.

The standardized pattern is:

```r
nodes    <- data.table(name = ..., type = ..., color = ...)
edgelist <- data.table(from = ..., to = ..., weight = ...)
graph    <- model_directed_graph(nodes, edgelist)   # or model_bipartite_graph / model_heterogeneous_graph
return(graph)
```

The choice of constructor depends on the graph type. Functions that produce developer-file or author-thread relationships use `model_bipartite_graph`. Functions that produce file-to-file or function-to-function dependency networks use `model_directed_graph`. Functions that combine multiple node and edge types use `model_heterogeneous_graph`.

Graph export functions such as `graph_to_dsmj` and graph projection functions such as `bipartite_graph_projection` and `temporal_graph_projection` use S3 dispatch on the class of the returned graph object. Because every transform function now returns a properly typed graph object, these downstream functions can dispatch correctly without any caller-side workarounds.

The `transform_reply_to_bipartite_network` function in `reply.R` covers the GitHub, Jira, and Mail pipelines since none of those have their own transform functions and all route through it. A `source` parameter was also added to this function, which stores the data source label as a column in both the nodes table and the edgelist. This allows downstream functions like engagement metrics to operate on reply objects from any platform without needing to know where the data came from, while still giving callers the option to filter by source if needed.

### Transform Functions

The following checklist shows the refactoring status of all `transform_*` functions across the codebase:

**R/git.R**
- [x] `transform_gitlog_to_temporal_network`
- [x] `transform_gitlog_to_bipartite_network`
- [x] `transform_gitlog_to_entity_bipartite_network`
- [x] `transform_gitlog_to_entity_temporal_network`
- [x] `transform_commit_message_id_to_network`
- [x] `transform_gitlog_to_heterogeneous_network`

**R/reply.R**
- [x] `transform_reply_to_bipartite_network` (covers GitHub, Jira, and Mail pipelines)

**R/src.R**
- [x] `transform_understand_dependencies_to_network`
- [x] `transform_dependencies_to_network`
- [x] `transform_r_dependencies_to_network`

**R/dv8.R**
- [x] `transform_dependencies_to_sdsmj`
- [x] `transform_gitlog_to_hdsmj`
- [x] `transform_temporal_gitlog_to_adsmj`

**R/vulnerabilities.R**
- [x] `transform_cve_cwe_file_to_network`

#### All Image Comparisons showcase Master Branch (left/top) vs Dev Branch (right/bottom)

- [x] `vignettes/gitlog_showcase.Rmd` — `transform_gitlog_to_bipartite_network`, `bipartite_graph_projection`, `transform_gitlog_to_temporal_network`

Image
Image
Image
Image
Image
Image
Image
Image
Image

- [x] `vignettes/gitlog_entity_showcase.Rmd` — `transform_gitlog_to_entity_bipartite_network`, `bipartite_graph_projection`, `transform_gitlog_to_entity_temporal_network`

Image
Image
Image
Image
Image
Image
Image

- [x] `vignettes/community_detection_showcase.Rmd` — `transform_gitlog_to_bipartite_network`, `bipartite_graph_projection`, `transform_gitlog_to_temporal_network`, `transform_dependencies_to_network`

Image
Image
Image
Image
Image

The two branches assign different **integer IDs** to nodes before running OSLOM.

OSLOM uses those IDs internally with a random seed. Same seed + different IDs = different community assignments and colors.

The ID assignment comes from `mapping_names <- unique(c(edgelist$from, edgelist$to))` in `community_oslom` — the order nodes appear in that list determines their ID. Because the S3 refactoring changed how the node table is built internally (all-from-nodes first then all-to-nodes, vs interleaved), the projected edgelist has a slightly different row order, which changes the `mapping_names` sequence, which changes the IDs fed to OSLOM.

**The weights are correct. The graph structure is correct. Only the node ID ordering going into OSLOM differs**, causing different (but equally valid) community partitions.

Image
Image

- [x] `vignettes/social_smell_showcase.Rmd` — `transform_gitlog_to_bipartite_network`, `bipartite_graph_projection`, `transform_reply_to_bipartite_network`

Image

- [x] `vignettes/issue_social_smell_showcase.Rmd` — `transform_gitlog_to_bipartite_network`, `bipartite_graph_projection`, `transform_reply_to_bipartite_network`, `transform_commit_message_id_to_network`

Image

- [x] `vignettes/causal_flaws.Rmd` — `transform_gitlog_to_bipartite_network`, `bipartite_graph_projection`, `transform_reply_to_bipartite_network`, `transform_gitlog_to_hdsmj`, `transform_dependencies_to_sdsmj`

Image
Image
Image

- [x] `vignettes/motif_analysis.Rmd` — `transform_gitlog_to_bipartite_network`, `bipartite_graph_projection`, `transform_reply_to_bipartite_network`, `transform_r_dependencies_to_network`

Image
Image

The `motif_analysis.Rmd` square motif section parses `../R` (the kaiaulu R source folder itself) to build a file dependency network. Because the dev branch adds new R source files and functions as part of issue #361, the parsed file dependency graph has one more edge than master's (30 vs 29). This changes the combined `file_git_reply_network` edge count (128 vs 127), which causes `visIgraph(randomSeed=1)` to produce a different force-directed layout — the same seed initializes differently when the graph topology changes.

The layout difference is expected and correct. It reflects that the dev branch contains additional R code, not a regression in the network construction logic.

Image

- [x] `vignettes/reply_communication_showcase.Rmd` — `transform_reply_to_bipartite_network`

Image

- [x] `vignettes/gitlog_vulnerabilities_showcase.Rmd` — `transform_commit_message_id_to_network`, `transform_cve_cwe_file_to_network`

Image
Image

- [x] `vignettes/bug_count.Rmd` — `transform_commit_message_id_to_network`

Image
Image

- [x] `vignettes/dv8_showcase.Rmd` — `transform_gitlog_to_hdsmj`, `transform_dependencies_to_sdsmj`

Image
Image
Image
Image

- [x] `vignettes/depends_showcase.Rmd` — `transform_dependencies_to_network`

Image

- [x] `vignettes/understand_showcase.Rmd` — `transform_understand_dependencies_to_network`

Image
Image
Image
Image

- [x] `vignettes/kaiaulu_architecture.Rmd` — `transform_r_dependencies_to_network`

Image
Image

Differences found in these images are due to the Kaiaulu Architecture notebook running on the current branch of Kaiaulu. The development branch includes new functions, causing differences in the notebook from Master to Dev.

## Graph to DSMJ Export File Test

[kaiaulu-bdsm.json](https://github.com/user-attachments/files/26815876/kaiaulu-bdsm.json)

## Tool Comparisons

**NetworkX, SNAP, and igraph**

NetworkX and SNAP both handle data loading and graph construction in a single step where you pass in raw data and get a graph object back with no separation between the two stages. igraph follows the same pattern in R where constructor functions like `graph_from_data_frame()` consume a data frame and return a graph object all at once.

igraph is the closest in similarity to Kaiaulu since it also uses R's S3 class system. The key difference is that igraph uses a single `igraph` class for all graph types and stores properties like directionality and bipartite membership as flags and vertex attributes on the same object. Kaiaulu instead uses separate S3 classes (`bipartite_graph`, `heterogeneous_graph`, `directed_graph`) which makes the graph type explicit and allows S3 dispatch to automatically route projection and export functions to the correct method. igraph supports bipartite graphs as a first-class feature but treats heterogeneous graphs by convention only where users manually set a type attribute and filter nodes and edges themselves.

**git2net and coronet**

These are the most direct comparisons since both libraries transform software engineering data into networks. git2net separates parsing from graph construction by persisting all extracted git data to a SQLite database first since mining a repository is expensive and should only be done once. coronet takes a more structured approach with a four-part architecture where configuration, data loading, network configuration, and graph construction are each handled by separate classes.

Kaiaulu currently sits between these two approaches. The transform functions do separate data preparation from graph construction but there is no persistent intermediate layer like a database or a formal data class connecting the two. The advantage of this is simplicity. Where Kaiaulu stands apart from both libraries is heterogeneous graph support. Neither git2net nor coronet provides a native heterogeneous graph type. Kaiaulu's `model_heterogeneous_graph` with explicit string node and edge types is a more expressive abstraction for multi-type networks that neither library currently offers.

**Quanteda**

Quanteda is the closest in similarity to where Kaiaulu's S3 system is heading. Quanteda defines a strict pipeline of S3 classes (`corpus` → `tokens` → `dfm`) where each class has its own dedicated set of functions following a `class_verb()` naming convention. This makes the API predictable and self-documenting since users always know what functions exist for a given object type. The core takeaway is that consistent naming conventions and per-class function sets can make a system intuitive and maintainable without requiring more complex class systems like S4 or R6.

Kaiaulu's `model_*` constructors and `bipartite_graph_projection` S3 generic are already moving in this direction.

**Codeface**

Codeface stores everything in a MySQL database and runs from the command line with no library API. Any analysis means querying MySQL directly or using a Shiny web app. It has not been maintained since 2022. The big takeaway is that a heavy setup creates a barrier. Having to install Vagrant, MySQL, and a web server just to run an analysis is a lot, and Kaiaulu avoids all of that by keeping things in memory.

**SmartSHARK**

SmartSHARK pushes separation of concerns the furthest out of all of these tools. Each data source is its own separate command-line tool, and they all write into a shared MongoDB database using a shared data format called pycoSHARK. The way SmartSHARK splits things by data source at the repo level is basically what Kaiaulu does at the file level with R/jira.R, R/github.R, and so on. It shows that the separation is the right call. It also shows what happens when you take it too far. Setting up a full SmartSHARK environment means deploying MongoDB, a Django web server, a background worker, and configuring every plugin on its own, which is way more than most researchers want to deal with.

**PyDriller**

PyDriller is a Python library that gives you a clean API for reading Git repositories. You create a Repository object, iterate over commits, and each commit gives you file changes, diffs, blame data, and complexity numbers. There is no command-line tool, and nothing gets saved to disk. The API is simple and easy to use. The downside is that every time you run an analysis, it has to re-read the whole repository, which is slow for big projects. Compared to Kaiaulu wrapping Perceval as a command-line tool, PyDriller shows what it looks like to build a proper library for Git data. Kaiaulu goes through Perceval, which adds a step, but it covers a lot more data sources than just Git.

**Perceval**

Perceval supports over 40 different data sources, including Git, GitHub, Jira, Bugzilla, mailing lists, and CI systems. It outputs one JSON object per item and streams it to the console. Kaiaulu uses it as a command-line dependency for parse_gitlog, so Kaiaulu runs the perceval program and reads its output rather than using it as a library. Perceval choosing to be a command-line tool instead of a library is the opposite of what PyDriller does. For Kaiaulu, this means the Git data collection part is stable and does not need to be maintained, but it also means it cannot be changed from within R. It is a clean split between collecting data and analyzing it, at the cost of needing Perceval installed.

jseto808:
### Summary

This PR standardizes the graph construction and transform APIs across all network types in Kaiaulu. It replaces the monolithic `model_directed_graph()` constructor with two purpose-built constructors (`model_unimodal_graph()`, `model_multimodal_graph()`), introduces S3 dispatch for projection and export functions, refactors all transform functions to explicitly build node and edgelist tables before passing them to constructors, and extends DV8 integration to support hierarchical clustering of author communication networks.

### Changes

**New graph model constructors (`R/graph.R`)**
- `model_unimodal_graph()` — replaces `model_directed_graph()` for single-node-type graphs
- `model_multimodal_graph()` — replaces `model_directed_graph()` for bipartite and multimodal graphs; supports `is_bipartite` flag and multiple edge types via an `edge_type` column
- `get_bidirected_edges()` — utility to extract edges where the reverse also exists
- `subset_bipartite_from_multimodal()` — extracts a bipartite subgraph from a multimodal graph by edge type

**S3 dispatch for projections and export (`R/graph.R`)**
- `bipartite_graph_projection()` — now dispatches on `is_bipartite` and `multimodal`; bipartite graphs embedded in a multimodal class are correctly routed
- `temporal_graph_projection()` — same dispatch pattern; validates `timestamp_column` presence; supports `edge_type` parameter to slice non-bipartite multimodal graphs before projection
- `graph_to_dsmj()` — now dispatches on `unimodal` and `multimodal`; infers direction from graph S3 class instead of an explicit `is_directed` parameter
- community_oslom() — now dispatches on unimodal and multimodal; multimodal graphs are routed through the unimodal method

**Refactored transform functions**
- `R/git.R`: `transform_gitlog_to_temporal_network`, `transform_gitlog_to_bipartite_network`, `transform_gitlog_to_entity_bipartite_network`, `transform_gitlog_to_entity_temporal_network`, `transform_commit_message_id_to_network` — all migrated to new constructors with explicit node/edgelist construction; weight semantics preserved from master (`weight = .N` for bipartite, `weight = 1L` per event for temporal, `n_lines_changed` for entity temporal)
- `R/git.R` (new): `transform_gitlog_to_multimodal_network` — constructs a three-edge-type multimodal graph (authored / changed / modified) from a parsed git log
- `R/src.R`: `transform_dependencies_to_network`, `transform_understand_dependencies_to_network`, `transform_r_dependencies_to_network` — migrated to `model_unimodal_graph()`
- `R/reply.R`: `transform_reply_to_bipartite_network` — migrated to `model_multimodal_graph()`
- `R/vulnerabilities.R`: `transform_cve_cwe_file_to_network` — migrated to `model_unimodal_graph()`
- `R/dv8.R`: `transform_dependencies_to_sdsmj`, `transform_gitlog_to_hdsmj`, `transform_temporal_gitlog_to_adsmj` — updated to use new constructors and S3-dispatched `graph_to_dsmj()`

**DV8 hierarchical clustering extensions (`R/dv8.R`)**
- `parse_dv8_clusters()` — rewritten to traverse the full cluster hierarchy recursively, emitting one row per leaf per layer; supports recursive DR Space output
- `dv8_mdsmb_to_hierclsxb()` — argument handling fixed; `recursive` parameter added

**New vignettes**
- `dv8_author_communication_showcase.Rmd` — end-to-end pipeline from mbox data to DV8 DR Space clustering of an author communication network
- `radio_silence_showcase.Rmd` — radio silence social smell with both OSLOM and DV8 DR Space per-layer community detection
- custom_graph_weight_showcase.Rmd — demonstrates custom edge weights in transform functions, using sentiment labels as an example; shows how to visualize the resulting graph with colored edges and weight labels  

**Bug fix**
- `transform_gitlog_to_entity_bipartite_network` committer-entity mode incorrectly used `from=author` instead of `from=committer`

### Test plan

Knit notebooks in the `master` branch and compare to knitted notebooks in this PR:

- [x] `gitlog_showcase.Rmd`
- [x] `gitlog_entity_showcase.Rmd`
- [x] `gitlog_vulnerabilities_showcase.Rmd`
- [x] `depends_showcase.Rmd`
- [x] `understand_showcase.Rmd`
- [x] `dv8_showcase.Rmd`
- [x] `custom_graph_weight_showcase.Rmd`
- [x] `dv8_author_communication_showcase.Rmd`
- [x] `radio_silence_showcase.Rmd`
- [x] `community_detection_showcase.Rmd`
- [x] `reply_communication_showcase.Rmd`
- [x] `social_smell_showcase.Rmd`
- [x] `issue_social_smell_showcase.Rmd`
- [x] `bug_count.Rmd`
- [x] `causal_flaws.Rmd`
- [x] `motif_analysis.Rmd`
- [x] `kaiaulu_architecture.Rmd`


jseto808:
PR checks fail because of the changes made to `model_directed_graph` function. Some files still use the old parameters (reply.R and src.R).

carlosparadis:
@jseto808 any reason why you holding on changing them? Too many changes / waiting confirmation/ something else ?

jseto808:
> @jseto808 any reason why you holding on changing them? Too many changes / waiting confirmation/ something else ?

Hi @carlosparadis, apologies for the confusion. I had listed that as a note to myself on what still needed to be fixed since I didn’t have time to address it at the moment.

R/git.R
```diff
@@ -938,48 +1000,37 @@ transform_gitlog_to_entity_bipartite_network <- function(project_git_entity, mod
...
+    from_color <- "black"
+  } else {
+    from_col   <- "committer_name_email"
+    dt_col     <- "committer_datetimetz"
+    from_color <- "#bed7be"
   }
-  git_graph[["edgelist"]] <- git_graph[["edgelist"]][,.(from,to,weight,datetimetz)]
-
-  temporal_projection <- temporal_graph_projection(git_graph,mode=TRUE,timestamp_column ="datetimetz",
+  to_col   <- "entity_definition_name"
```
phuong808: Looking at the old code being removed, the actual column being accessed was ```entity_definition_name```, but in ```transform_gitlog_to_entity_bipartite_network``` (also in this PR) the equivalent column is just ```"entity"```. If both functions operate on the same input from ```parse_gitlog_entity```, these should be consistent. Worth double-checking what column name ```parse_gitlog_entity``` actually returns — if it's ```"entity"```, this will silently produce NULL nodes.

R/git.R
```diff
@@ -990,20 +1041,134 @@ transform_gitlog_to_entity_temporal_network <- function(project_git_entity,mode
...
+    # All rows sharing the same commit_hash have the same timestamp so take first
+    # lines_added and lines_removed are stored as characters by parse_gitlog
+    edge_tables[["changed"]] <- project_git[,.(
+      weight     = sum(as.numeric(lines_added) + as.numeric(lines_removed), na.rm = TRUE),
+      type       = "changed",
+      color      = "#afe569",
+      datetimetz = author_datetimetz[1]
+    ), by = .(from = commit_hash, to = file_pathname)]
+  }
+  if(all(c("author_name_email","file_pathname") %in% node_cols) &
```
phuong808: ```&``` is vectorised — it works here since both sides are length-1 logicals, but ```&&``` is the correct scalar logical operator for ```if``` conditions in R. It's a minor style/correctness point but could produce unexpected behaviour with non-scalar inputs.

R/git.R
```diff
@@ -990,20 +1041,134 @@ transform_gitlog_to_entity_temporal_network <- function(project_git_entity,mode
...
-
-  # Keep only the edges which contain the commit message id
-
-  project_git <- project_git[!is.na(commit_message_id),.(commit_message_id,
-                                                         file_pathname)]
-
-  git_graph <- model_directed_graph(project_git[,.(from=commit_message_id,to=file_pathname)],
-                                    is_bipartite=TRUE,
-                                    color=c("#0052cc","#f4dbb5"))
+  project_git[, commit_message_id := stringi::stri_match_first_regex(commit_message, pattern = commit_message_id_regex)[, 1]]
```
phuong808, line 1079: ```:=``` modifies the caller's ```project_git``` data.table by reference — unexpected side effect. Suggestion: Add ```project_git <- copy(project_git)``` before this line, consistent with how other functions in this file guard against mutation.

R/graph.R
```diff
@@ -504,6 +670,7 @@ temporal_graph_projection <- function(graph,mode,weight_scheme_function = NULL,t
                                                .SDcols = c("to","weight","datetimetz")]
 
     if(nrow(graph[["edgelist"]] == 0)){
+      class(graph) <- c("directed_graph", "list")
       return(graph)
     }else{
       setnames(x = graph[["edgelist"]],
```
phuong808, line 707: The two early-return guards are inconsistent with each other. The first block correctly uses ```nrow(graph[["edgelist"]]) == 0``` with ```== 0``` outside ```nrow()```. The second block uses ```nrow(graph[["edgelist"]] == 0)``` with ```== 0``` inside ```nrow()```, which applies the comparison element-wise to the entire data.table and passes a logical matrix to ```nrow()``` — it will never correctly detect an empty edgelist. The second block should be changed to match the first: ```nrow(graph[["edgelist"]]) ==```

R/graph.R
```diff
@@ -102,170 +138,299 @@ graph_to_dsmj <- function(graph, dsmj_path, dsmj_name, is_directed, is_sorted=FA
...
+  if(nrow(slice_edges) == 0){
+    stop(paste0("No edges found with edge_type '", edge_type, "'."))
+  }
+
+  if(!(timestamp_column %in% names(slice_edges))){
+    stop(paste0("timestamp_column '", timestamp_column, "' not found in edges of type '", edge_type, "'."))
+  }
+
+  # Identify the two node types involved in this edge type
+  from_type <- unique(graph[["nodes"]][name %in% slice_edges$from]$type)
```
phuong808: Same with line 336

R/graph.R
```diff
@@ -102,170 +138,299 @@ graph_to_dsmj <- function(graph, dsmj_path, dsmj_name, is_directed, is_sorted=FA
...
+bipartite_graph_projection.heterogeneous_graph <- function(graph, mode, edge_type, weight_scheme_function=NULL, ...){
+  # Extract edgelist for the requested edge type, keeping only from/to/weight
+  slice_edges <- graph[["edgelist"]][type == edge_type, .(from, to, weight)]
+
+  if(nrow(slice_edges) == 0){
+    stop(paste0("No edges found with edge_type '", edge_type, "'."))
+  }
+
+  # Identify the two node types involved in this edge type
+  from_type <- unique(graph[["nodes"]][name %in% slice_edges$from]$type)
```
phuong808: If the heterogeneous graph has nodes of multiple types appearing on the ```from``` side of this edge type, ```from_type``` will be length > 1. The downstream ```slice_nodes[, type := (type == from_type)]``` will silently use only the first element, producing wrong ```TRUE/FALSE``` type recoding without any error. Suggest adding an explicit guard:
``` r
if(length(from_type) != 1 || length(to_type) != 1){
  stop("edge_type '", edge_type, "' connects more than two node types — cannot project.")
}
```

phuong808 (approved):

R/git.R
jseto808: The `entity` column in `parse_gitlog_entity` is an alias added explicitly for naming consistency with `parse_gitlog`. Both `entity` and `entity_definition_name` hold the same value. The old temporal function was using `entity_definition_name` while the bipartite function was using `entity`, so we standardized both to `entity`. The redundancy between the two columns is a separate issue worth cleaning up in the parser, but that is out of scope for this PR.

carlosparadis:
Interesting, I just realized this was made as a branch out of a fork, rather than a branch within Kaiaulu (the write permission was so you could branch off Kaiaulu!). But not a big deal 👍

vignettes/radio_silence_showcase.Rmd
```diff
@@ -0,0 +1,217 @@
...
+```{r}
+project_github_replies <- parse_github_replies(
+  issues_json_folder_path        = github_issue_path,
+  pull_requests_json_folder_path = github_pull_request_path,
+  comments_json_folder_path      = github_reply_path,
+  commit_json_folder_path        = github_commit_path,
+  pr_comments_json_folder_path   = github_pr_comments_path
+)
+nrow(project_github_replies)
+kable(head(project_github_replies[, .(reply_from, reply_subject)]))
```
carlosparadis: kable() functions often break format in more complex data. This should use gt(). There is an example in Kaiaulu. It is also a much more powerful table composition library, and it only needs one line replacement. Think you could replace them? Likely can be done with a ctrl+f and replace.

vignettes/radio_silence_showcase.Rmd
carlosparadis, line 1: See my commit fixing kaiaulu.yml for the correction I made here, no action item. Just FYI.

vignettes/custom_graph_weight_showcase.Rmd
```diff
@@ -0,0 +1,199 @@
...
+This notebook demonstrates how to assign custom numeric values as edge weights when transforming raw data into a graph object. Specifically, it uses sentiment polarity scores from GitHub issue comments as edge weights in a bipartite author-to-issue network, and visualizes the resulting network with edges colored by sentiment.
+
+Polarity integers (`0` = neutral, `1` = positive, `2` = negative) are remapped to (`0`, `1`, `-1`) before being passed to `transform_reply_to_bipartite_network()` via the `weight` parameter. The `weight_agg` parameter controls how multiple comments between the same author and issue are aggregated — defaulting to `mean`.
+
+# Load Data
+
+The input is a CSV of Kaiaulu GitHub issue comments with predicted sentiment polarity labels.
+
+```{r}
+dt <- fread("https://github.com/user-attachments/files/27417102/kaiaulu_predictions.csv")
```
carlosparadis, line 40: Bravo! You circumvented the data dependency that would make me unable to test this notebook without running the sentiment analysis Notebook, by referencing it directly to the issue attachment. Really well done. It is such a simple solution but this is the first time I've seen being used with a data.table fread in the project. +1

vignettes/custom_graph_weight_showcase.Rmd
carlosparadis, line 1: Could execute by just pressing run all. Flawless!

vignettes/dv8_showcase.Rmd
```diff
@@ -276,13 +276,14 @@ hierclsxj_path <- dv8_clsxb_to_clsxj(dv8_path,
                                      clsxj_path = file.path(project_path,paste0(project_name,
                                                                                 "-clsxj.json")))
 hierclsxj_table <- parse_dv8_clusters(hierclsxj_path)
-kable(head(hierclsxj_table))
+top_hierclsxj_table <- hierclsxj_table[!grepl("/", layer)]
+kable(head(top_hierclsxj_table))
```
carlosparadis, line 280: @jseto808 why was this modified? I feel I am missing something.

carlosparadis (changes requested):
@jseto808 

In order for me to do the final test, which runs all notebooks in Kaiaulu using your new api, I need to list the functions you added and deleted in kaiaulu/_pkgdown.yml. This is where I define the modules you see in the itm0 docs. 

Could you version that file in your PR, add the functions you added and remove those from there you deleted for me? 

I am done making commits here for now. 

Remember that you need to git checkout to this branch on your local machine, then git pull my commits **before** making changes. If you try to edit before doing that, you will enter a merge conflict which is way more annoying to deal with. 

Please ping me here when you can pass the torch so we don't enter commit conflict. 

This is really, really good. I could run all notebooks very easily by literally git pulling and pressing run all. 

I see you also respected the relative paths, folder conventions, and did not commit the tools.yml with your local config (a very common mistake).

I will take a closer look on the transforms after I can use the pkgdown to run all notebooks, but I imagine any major change here would be too late to patch at this point. Fingers crossed it works across all notebooks!

vignettes/dv8_showcase.Rmd
jseto808: @carlosparadis `parse_dv8_clusters` now returns one row per file per layer. Since the flaws map only needs one cluster assignment per file, the fix keeps only the top-level layer rows before merging to avoid duplicates.

jseto808:
@carlosparadis I updated the `_pkgdown.yml` with the functions I added/deleted

carlosparadis:
Acknowledged. Will review after I take a pass on Phuong's code. Can you issue a code review request for me on the PR?

carlosparadis:
@jseto808 I was able to finish running the code through all notebooks. I have not inspected the outputs yet, but here's what I would say breaks the code and prevents this PR from merge. 

FYI you can test it yourself by doing `pkgdown::build_articles()` and having the article open or using article and specify the article of interest, although you of course may be blocked depending on the third party tool.

# vignettes/understand_showcase.Rmd

Both understand transforms are breaking. 

https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/vignettes/understand_showcase.Rmd#L108-L136

# vignettes/community_detection_showcase.Rmd

https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/vignettes/community_detection_showcase.Rmd#L221-L228

This recolor function now breaks.

# vignettes/causal_flaws.Rmd

https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/vignettes/causal_flaws.Rmd#L483-L489

The chunk 35 keeps breaking when I use pkgdown, but can run and compile on knitr. Not sure what is going on.

# That's it.

Surprisingly that's all that crashed across **ALL** Kaiaulu notebooks when compiling itm0 docs. I also sent you via email the drive url of the generated docs. You should need to open the index.html file and you can browse locally.

As I said before, there is little to no time. If it is an easy fix, go for it. If not, I'd appreciate if you could add some pointers here for Fall'26 to help me patch this up. I also need to of course look through the output notebooks to see if the code didn't run and had NAs where it used to be results.

Note you will not see causal_flaws.Rmd there since I had to remove it from compiling altogether. The rest will show up but the code block that breaks i modified to eval = false so there will be no output for them.

Signing off for now, just wanted to handle this over to you the earliest I could.

carlosparadis (changes requested):
see https://github.com/sailuh/kaiaulu/pull/375#issuecomment-4421703806

jseto808:
Hi @carlosparadis,

Here is a report of the three breaking notebooks, what caused each failure, and what was fixed.

## understand_showcase.Rmd

**What broke:** Both `transform_understand_dependencies_to_network` calls failed immediately with "unused argument" because the parameter was passed as `parsed =` but the function signature on this branch uses `understand_parsed`.

**Fix:** Changed both call sites in the vignette:
```r
# Before
transform_understand_dependencies_to_network(parsed = file_dependencies, ...)
transform_understand_dependencies_to_network(parsed = class_dependencies, ...)

# After
transform_understand_dependencies_to_network(understand_parsed = file_dependencies, ...)
transform_understand_dependencies_to_network(understand_parsed = class_dependencies, ...)
```

## community_detection_showcase.Rmd

**What broke:** Reported as the `recolor_network_by_community` call failing. I was able to successfully knit and run pkgdown on the notebook with no errors. Additional inspection of the article produced from pkgdown build showed no issues as well.

**Fix:** No code change.

## causal_flaws.Rmd

**What broke:**  Column mismatch in `rbind`:** When combining `git_network`, `reply_network`, and `file_network` edgelists, the `rbind` failed with "Item 3 has 3 columns, inconsistent with item 1 which has 4 columns." On this branch, the new transform functions (`transform_gitlog_to_bipartite_network`, `transform_reply_to_bipartite_network`) now include a `direction` column in their edgelists. The `file_network` edgelist was built manually in the vignette without that column.

**Fix:** Added `direction` to the manual edgelist construction:
```r
edgelist <- file_network[["edgelist"]][,.(from=src_filepath, to=dest_filepath)]
edgelist$weight <- rowSums(...)
edgelist[, direction := "directed"]   # added
file_network[["edgelist"]] <- edgelist
```

carlosparadis:
Acknowledged. Will leave it running on the fixes and let you know how it goes, thanks!

carlosparadis:
# community_detection_showcase.Rmd

There is something going wrong with 


```r
project_commit_network <- transform_gitlog_to_bipartite_network(project_git_slice,
                                                                mode="commit-file")
```

I don't think it is happening to you because this may be tied to the dataset. The error is:

```
Warning: Item 1 has 0 rows but longest item has 1; filled with NAWarning: Item 1 has 0 rows but longest item has 1; filled with NA
```

I sent you the repo of APR I am using. You can use the apr config to run. Could you try to run with this to see if it blows up on your side?

carlosparadis:
@jseto808 

Disregard comment above. Problem was a slice that was emptying the table:

#project_git_slice <- project_git[author_datetimetz >= as.POSIXct("2015-01-01", format = "%Y-%m-%d",tz = "UTC") & author_datetimetz < as.POSIXct("2019-12-31", format = "%Y-%m-%d",tz = "UTC")]

project_git_slice <- project_git[author_datetimetz >= as.POSIXct("1999-01-01", format = "%Y-%m-%d",tz = "UTC") & author_datetimetz < as.POSIXct("2000-12-31", format = "%Y-%m-%d",tz = "UTC")]

That was causing memory errors. Commented led to a empty git slice that would then blow up on the recolor.

Testing causal flaws next. Understand now works.

carlosparadis:
Individually, I can now knit all notebooks! I am doing a run all to see how it goes.

carlosparadis (approved):
The last commit compiles all itm0 docs, meaning all notebooks currently execute start to end without error with the new added changes!

I still need to look at the docs of itm0 to sanity check the actual output and the logic before merging but this is as good as it gets with what time is left.

Issue #360, opened 2026-02-01
Implement Engagement Metrics


usradam:
# Purpose
## Engagement Communication
```r
engagement_communication <- function(datetimetz, user_name_email, quit_lag = 90)
```
The `engagement_communication` function checks the author (`user_name_email`) and applies the rolling window (`quit_lag`) of 90 days. For each window of 90 days, the function counts (using a function that counts so in data.table the .N will likely be needed) up the number of comments for each author and outputs that as a table as shown above.

## Engagement Sentiment
```r
engagement_sentiment <- function(datetimetz, user_name_email, polarity, quit_lag = 90, aggregate_func)
```
For the `engagement_sentiment`, it re-uses much of the logic as `engagement_communication` but it will instead apply an aggregate function to the sentiment (`polarity`) for each 90 day window (`quit_lag`) from the author (`user_name_email`). This way, we can see for each window, what an author's overall sentiment was. We are still taking into account Samantha's data, so then the string values that her data has e.g. `positive`, `neutral`, `negative` would have to be converted to the numerical representation first to then apply an aggregate function for the overall polarity for each author's window. The `aggregate_func` is passed as parameter. In R, you can pass both variables and function definitions as parameters. Therefore, if it is the mean() function, that is built in R, and the user just needs to type "mean". Same with "sd" for standard deviation.

# Process
#### Input Table
datetimetz | user_message | user_name_email
-- | -- | --
2011-01-01T11:39:45Z | "Why does it require a HashSet?\n" | Alice Wonderland <<alice@example.com>>
2024-01-10T11:39:45Z | "I am done with the refactor\n" | Alice Wonderland <<alice@example.com>>
2024-03-15T11:39:45Z | "I don't want to work on this anymore\n" | Alice Wonderland <<alice@example.com>>
2024-05-01T11:39:45Z | "Is this feature done?\n"| Alice Wonderland <<alice@example.com>>
2024-01-05T11:39:45Z | "Stop taking so long\n" | Bob Smith <<bob@example.com>>
2024-04-10T11:39:45Z | "What did you mean?\n" | Bob Smith <<bob@example.com>>

#### Output Table
datetimetz | user_name_email | message_count
-- | -- | --
2024-01-01T11:39:45Z | Alice Wonderland <<alice@example.com>> | 3
2024-04-01T11:39:45Z | Alice Wonderland <<alice@example.com>> | 1
2024-01-01T11:39:45Z | Bob Smith <<bob@example.com>> | 1
[...]

PR #362, opened 2026-02-18
Implementing Engagement Metrics

usradam:
# Purpose
## Engagement Communication
```r
engagement_communication <- function(datetimetz, user_name_email, quit_lag = 90)
```
The `engagement_communication` function checks the author (`user_name_email`) and applies the rolling window (`quit_lag`) of 90 days. For each window of 90 days, the function counts (using a function that counts so in data.table the .N will likely be needed) up the number of comments for each author and outputs that as a table as shown above.

## Engagement Sentiment
```r
engagement_sentiment <- function(datetimetz, user_name_email, polarity, quit_lag = 90, aggregate_func)
```
For the `engagement_sentiment`, it re-uses much of the logic as `engagement_communication` but it will instead apply an aggregate function to the sentiment (`polarity`) for each 90 day window (`quit_lag`) from the author (`user_name_email`). This way, we can see for each window, what an author's overall sentiment was. We are still taking into account Samantha's data, so then the string values that her data has e.g. `positive`, `neutral`, `negative` would have to be converted to the numerical representation first to then apply an aggregate function for the overall polarity for each author's window. The `aggregate_func` is passed as parameter. In R, you can pass both variables and function definitions as parameters. Therefore, if it is the mean() function, that is built in R, and the user just needs to type "mean". Same with "sd" for standard deviation.

# Process
#### Input Table
datetimetz | user_message | user_name_email
-- | -- | --
2011-01-01T11:39:45Z | "Why does it require a HashSet?\n" | Alice Wonderland <<alice@example.com>>
2024-01-10T11:39:45Z | "I am done with the refactor\n" | Alice Wonderland <<alice@example.com>>
2024-03-15T11:39:45Z | "I don't want to work on this anymore\n" | Alice Wonderland <<alice@example.com>>
2024-05-01T11:39:45Z | "Is this feature done?\n"| Alice Wonderland <<alice@example.com>>
2024-01-05T11:39:45Z | "Stop taking so long\n" | Bob Smith <<bob@example.com>>
2024-04-10T11:39:45Z | "What did you mean?\n" | Bob Smith <<bob@example.com>>

#### Output Table
datetimetz | user_name_email | message_count
-- | -- | --
2024-01-01T11:39:45Z | Alice Wonderland <<alice@example.com>> | 3
2024-04-01T11:39:45Z | Alice Wonderland <<alice@example.com>> | 1
2024-01-01T11:39:45Z | Bob Smith <<bob@example.com>> | 1
[...]

PR #380, opened 2026-03-19
Implementing Productivity Metrics

### LINKED ISSUE #377: Implement Productivity Metrics
# Purpose
Coming off [the engagement metrics](https://github.com/sailuh/kaiaulu/issues/360), we now want to measure developer productivity through commits and lines of code (LOC).

Per Rick's email:
> "Quitting is simply the low end of the engagement scale and rage is the high end of the sentiment/emotion scale. But in this study we are seeking a more nuanced understanding of the relationships. The hypothesis would be that lower sentiment/emotion scores will correlate with lower engagement and lower productivity."

# Process
- Have the productivity metrics for commits and churn be implemented as two functions called 1) `productivity_author_commits` and 2) `productivity_author_churn`
- Understand the input and output of these functions
- Create the function signatures, ensuring relevant parameters are included
- Both functions take into account the 90-day rolling window for its aggregation as we have looked at before with the engagement metrics

## Implementing Commits Productivity
For commits we can use `commit_hash` from the `parse_gitlog` data table to get the unique commits per `author_name_email`. I can then get a simple summation of the overall number of commits. The resulting output would be, for example, the following:
author_name_email | author_datetimetz | author_total_commits
-- | -- | --
Alice Wonderland <<alice@example.com>> | 2024-01-10T11:39:45Z | 5
Bob Smith <<bob@example.com>> | 2024-01-10T11:39:45Z | 3
Charlie Brown <<charlie@example.com>> | 2024-01-05T11:39:45Z | 9

## Implementing Author Churn
I looked at churn metrics that are already implemented including `metric_churn()`, `metric_file_churn()`, `metric_churn_per_commit_per_file()`, etc. and none at first glance take into account per developer churn. 

As such, I can expand upon the existing functions rather than re-do any of them by using [`metric_churn_per_commit_per_file`](https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/R/metric.R#L156) which takes into account the [`metric_churn`](https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/R/metric.R#L120) function. I see that `metric_churn`
defines churn as the following:
```r
churn <- lines_added + lines_removed
```
So no negative (-) number consideration as Carlos mentions. It treats `lines_added` and `lines_removed` as non-negative counts and defines churn as the summation of the two numbers. As such the example output data table for churn from above would instead look like the following:
author_name_email | author_datetimetz | lines_added | lines_removed | author_churn
-- | -- | -- | -- | --
Alice Wonderland <<alice@example.com>> | 2024-01-01T11:39:45Z | 500 | 50 | 550
Bob Smith <<bob@example.com>> | 2024-04-01T11:39:45Z | 200 | 400 | 600

Using `metric_churn_per_commit_per_file` I can pass in `project_git_log`, which is the data from `parse_gitlog`. Then I can define a variable to hold the developer churn data e.g. `author_churn` and then get the sum of the churn grouped by `author_name_email`. The resulting output would look like the table above.

# Task List
- [x] Iterate understanding of the input and output for the functions
- [x] Create function signatures
- [x] Implement functions

usradam:
# Purpose
Coming off [the engagement metrics](https://github.com/sailuh/kaiaulu/issues/360), we now want to measure developer productivity through commits and lines of code (LOC).

Per Rick's email:
> "Quitting is simply the low end of the engagement scale and rage is the high end of the sentiment/emotion scale. But in this study we are seeking a more nuanced understanding of the relationships. The hypothesis would be that lower sentiment/emotion scores will correlate with lower engagement and lower productivity."

# Process
- Have the productivity metrics for commits and churn be implemented as two functions called 1) `productivity_author_commits` and 2) `productivity_author_churn`
- Understand the input and output of these functions
- Create the function signatures, ensuring relevant parameters are included
- Both functions take into account the 90-day rolling window for its aggregation as we have looked at before with the engagement metrics

## Implementing Commits Productivity
For commits we can use `commit_hash` from the `parse_gitlog` data table to get the unique commits per `author_name_email`. I can then get a simple summation of the overall number of commits. The resulting output would be, for example, the following:
author_name_email | author_datetimetz | author_total_commits
-- | -- | --
Alice Wonderland <<alice@example.com>> | 2024-01-10T11:39:45Z | 5
Bob Smith <<bob@example.com>> | 2024-01-10T11:39:45Z | 3
Charlie Brown <<charlie@example.com>> | 2024-01-05T11:39:45Z | 9

## Implementing Author Churn
I looked at churn metrics that are already implemented including `metric_churn()`, `metric_file_churn()`, `metric_churn_per_commit_per_file()`, etc. and none at first glance take into account per developer churn. 

As such, I can expand upon the existing functions rather than re-do any of them by using [`metric_churn_per_commit_per_file`](https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/R/metric.R#L156) which takes into account the [`metric_churn`](https://github.com/sailuh/kaiaulu/blob/5bbdbb2834df02971e604cb20207f2d25e3acc56/R/metric.R#L120) function. I see that `metric_churn`
defines churn as the following:
```r
churn <- lines_added + lines_removed
```
So no negative (-) number consideration as Carlos mentions. It treats `lines_added` and `lines_removed` as non-negative counts and defines churn as the summation of the two numbers. As such the example output data table for churn from above would instead look like the following:
author_name_email | author_datetimetz | lines_added | lines_removed | author_churn
-- | -- | -- | -- | --
Alice Wonderland <<alice@example.com>> | 2024-01-01T11:39:45Z | 500 | 50 | 550
Bob Smith <<bob@example.com>> | 2024-04-01T11:39:45Z | 200 | 400 | 600

Using `metric_churn_per_commit_per_file` I can pass in `project_git_log`, which is the data from `parse_gitlog`. Then I can define a variable to hold the developer churn data e.g. `author_churn` and then get the sum of the churn grouped by `author_name_email`. The resulting output would look like the table above.

# Task List
- [x] Iterate understanding of the input and output for the functions
- [x] Create function signatures
- [x] Implement functions


R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+      days = -quit_lag,
+      tz = tz_val
+    )
+
+    # Find the index of the last value in author_datetimetz that is <= the window_start for each row
+    # +1L is used later to get the first value inside the window
+    start_idx <- findInterval(window_start, author_datetimetz)
+
+    # Compute the number of unique commits for each row based on rolling window
+    author_total_commits <- vapply(seq_len(.N), function(i) {
```
phuong808: This  ```vapply``` loop is O(n²) per author — for each row it rebuilds the window index and re-aggregates from scratch. On authors with large commit histories this will degrade significantly. Consider replacing with ```slider::slide_index_dbl() / slide_index_int()``` which handles rolling windows over an index (like a datetime) natively and is fully vectorized. For example, ```slider::slide_index_int(commit_hash, author_datetimetz, ~uniqueN(.x), .before = days(quit_lag))```.

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+      idx <- seq.int(start_idx[i] + 1L, i)
```
phuong808: If no commits fall within the rolling window, ```start_idx[i] + 1L``` will be greater than ```i```, causing ```seq.int``` to silently return a descending sequence, for example(```seq.int(3, 1)``` → ```3 2 1```) instead of an empty vector. This means the aggregation picks up rows outside the intended window and returns a wrong result rather than ```0```. Suggest guarding with: ```idx <- if (start_idx[i] + 1L > i) integer(0) else seq.int(start_idx[i] + 1L, i)```.

phuong808 (changes requested):
Two issues flagged across ```productivity_author_commits``` and ```productivity_author_churn``` function. First, a silent correctness bug in the ```seq.int``` empty window edge case, where authors with sparse commit history, for example when there are no commits in the lookback window, this will produce inflated rolling counts/churn due to the descending sequence fallback. Second, a performance concern with the ```vapply``` rolling loop which is O(n²) per author and will become a bottleneck as git history grows. Since both issues share the same pattern across both functions, any fix should be applied consistently.

R/metric.R
usradam: For now, it is not optimal to introduce outside libraries like `slider` to Kaiaulu due to licensing logistics. You can refer to [this comment](https://github.com/sailuh/kaiaulu/issues/360#issuecomment-3902374088) Carlos made under the "third step" header in regards to introducing new dependencies into Kaiaulu.

R/metric.R
usradam: I don't think this would be a problem. For each row `i` I do `window_start[i] = author_datetimetz[i] - quit_lag days`, and with a `quit_lag` of 90 days it guarantees `window_start[i] < author_datetimetz[i]`.

After there's `start_idx <- findInterval(window_start, author_datetimetz)` i.e. the index of the last value in `author_datetimetz` that is `<= window_start[i]`. Since `author_datetimetz[i]` is greater than `window_start[i]` then at most `start_idx` returns `i-1`.

phuong808 (approved):

R/metric.R
carlosparadis: Hi,

`findInterval` breaks convention in R and Kaiaulu: We do not use camel case. Please ensure the entire team stays consistent to Kaiaulu conventions. Thanks!

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+  result <- dt[, {
+    window_start <- stringi::stri_datetime_add(
+      author_datetimetz,
+      days = -quit_lag,
+      tz = tz_val
+    )
+
+    # Find the index of the last value in author_datetimetz that is <= the window_start for each row
+    # +1L is used later to get the first value inside the window
+    start_idx <- findInterval(window_start, author_datetimetz)
```
carlosparadis: why this is camel case when the others are all consistent to lower case with `_`?

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+#'
+#' @param project_git_log a parsed git log obtained from \code{\link{parse_gitlog}}
+#' @param quit_lag The number of days to look back for the rolling window (90 day default)
+#' @return A three column data.table of the form author_name_email | author_datetimetz | author_total_commits.
+#' For each row (each timestamp in author_datetimetz), author_total_commits is the number of unique 
+#' commits by that author in the preceding 90 days ending at that timestamp.
+#' @export
+#' @family metrics
+#' @seealso \code{\link{parse_gitlog}}
+productivity_author_commits <- function(project_git_log, quit_lag = 90) {
```
carlosparadis: the parameter name quit_lag does not make sense anymore, Who is quiting here? the variable should just be called lag. Careful with the wording to avoid misleading the user!

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
                                          pattern = commit_message_id_regex)
   return(length(is_match[is_match]))
 }
+
+#' Productivity Author Commits
+#'
+#' Counts the unique number of commits per author in a rolling window based on 
+#' the git log.
+#'
+#' @param project_git_log a parsed git log obtained from \code{\link{parse_gitlog}}
```
carlosparadis, line 190: why is the parameter called `project_git_log` wheni the function calls it gitlog? Stay consistent! @jseto808 you should be picking these up as architect reviewer.

R/metric.R
carlosparadis: I rarely use vapply in Kaiaulu if at all. Could this be a sapply or lapply + rbindlist instead?

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+#' @export
+#' @family metrics
+#' @seealso \code{\link{parse_gitlog}}
+productivity_author_commits <- function(project_git_log, quit_lag = 90) {
+  author_name_email <- author_datetimetz <- commit_hash <- NULL
+
+  # Determine timezone from the author_datetimetz column in project_git_log
+  tz_val <- attr(project_git_log$author_datetimetz, "tzone")
+  if (is.null(tz_val)) {
+    tz_val <- Sys.timezone()
```
carlosparadis: No idea why sys.timezone is being used here. Nothing in the code should be dependent on the timezone local to the person running the code. This would lead to inconsistent results depending on who runs the code. Can you clarify why this is here?

R/metric.R
carlosparadis: Edit: I just realize this is not your implementation. If so, why is :: not being used as it were on the other functions? I am also curious if data.table can't implement what you are doing here. I'd like to avoid depending on base R as much as possible as it is far more inefficient than data.table dependencies which is implemented in C and C++ underneath.

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+#' @param project_git_log a parsed git log obtained from \code{\link{parse_gitlog}}
+#' @param quit_lag The number of days to look back for the rolling window (90 day default)
+#' @return a five column data.table of the form author_name_email | author_datetimetz | lines_added | 
+#' lines_removed | author_churn. 
+#' At each timestamp, the lines_added, lines_removed, and author_churn are the totals across all rows 
+#' for that author within the prior 90 days ending at that timestamp.
+#' @export
+#' @family metrics
+#' @seealso \code{\link{parse_gitlog}} to obtain additions and deletions from gitlog
+productivity_author_churn <- function(project_git_log, quit_lag = 90) {
```
carlosparadis: Same issue with `quit_lag`. Rename to lag and ensure others are as well.

R/metric.R
```diff
@@ -181,6 +181,137 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+    .(author_name_email = author_name_email,
+      author_datetimetz = author_datetimetz,
+      lines_added = lines_added_roll,
+      lines_removed = lines_removed_roll,
+      author_churn = churn_roll)
+
+  }, by = .(author_name_email)]
+
+  return(result[])
+}
```
carlosparadis, line 291: No code in Kaiaulu has this giant code block loaded dynamically within dt. This is not react. Please revise to stay consistent to how Kaiaulu and R code in general is written.

carlosparadis (changes requested):
I am very concerned with how inconsistent the code is written here. The architect should be ensuring the code does not look inconsistent to the rest of the codebase. @jseto808

carlosparadis:
These are not the only code review changes but what I pointed above should already require substantial revision. Please ensure the rest of the team is on the same page here, especially with the way the code is written.

R/metric.R
usradam: @carlosparadis I got it from your suggestion from [this comment](https://github.com/sailuh/kaiaulu/issues/377#issuecomment-4061444873). You didn't say anything so I assumed it was fine. In the "Follow-Up and Other Considerations" I noted the inconsistency in Kaiaulu in the names for this, so I went with what I stated.

R/metric.R
usradam: @carlosparadis Yes it is a base function in R which is in camel case. I used it after [a note from Phuong](https://github.com/sailuh/kaiaulu/pull/362#discussion_r2898585766) from my engagement metrics PR that my initial implementation could be slow for large datasets. I used a nested data.table approach at first so I looked around and thought that I could use findInterval since it does an interval search that gets O (n log N) complexity.

I can go back to my engagement metrics PR and this one and do the original data.table approach if that is better.

R/metric.R
usradam: @carlosparadis Yeah I was copying code from my engagement metrics so I missed this. I will fix it.

R/metric.R
carlosparadis: Please use the original. You have to account for maintainability the most in this project. Simple code, less bugs, faster onboarding!

R/metric.R
usradam: @carlosparadis I could use sapply instead if we want to stick as much to Kaiaulu style as possible. I used vapply because I was Googling and read that it offers more type safety and is faster than sapply.

I am reverting back to my original approach instead now so there is now no need to worry about vapply, sapply, or lapply + rbindlist.

R/metric.R
carlosparadis: There is someone in the other group looking out for these inconsistencies (john) so triple check the parameters.

R/metric.R
usradam: @carlosparadis This was just a check for null/missing values after extracting the timezone from the datetimetz variable. I will delete it so it does not assume a timezone local to the person running it.

R/metric.R
usradam: I'll get on this and make sure that this is more modular.

usradam:
@carlosparadis Thanks for the checks. I am gonna go back and do the revisions based on what we both just commented on above. I will also go back to the engagement metrics PR to ensure the same consistency.

R/metric.R
usradam: I will just stick with the name `project_git_log` for now since there are the are all the other inconsistencies in Kaiaulu that already need to be checked. `project_git_log` is the name that appears in git.R which is where `parse_gitlog` resides, so that name fits best for now.

carlosparadis:
@usradam Yes, this is fine to have you notebook with all metrics but the sentiment notebook one.

usradam:
@carlosparadis Thank you, got it. I'll go ahead with the planned tasks mentioned in [my issue comment](https://github.com/sailuh/kaiaulu/issues/377#issuecomment-4292374666). I appreciate the help and clarification.

vignettes/productivity_communication_metrics.Rmd
```diff
@@ -0,0 +1,250 @@
...
+    number_sections: true
+vignette: >
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteIndexEntry{Exploring the Productivity and Communication Metrics}
+  %\VignetteEncoding{UTF-8}
+---
+
+# Introduction
+
+This notebook presents some of the functionality of the following functions from `metric.R`: `productivity_author_commits`, `productivity_author_churn`, and `jira_developer_communication_count`.
```
jseto808, line 15: `jira_developer_communication_count` doesn't exist in `metric.R` within this PR

R/metric.R
```diff
@@ -181,6 +181,115 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
                                          pattern = commit_message_id_regex)
   return(length(is_match[is_match]))
 }
+
+#' Productivity Author Commits Metric
+#'
+#' Counts the unique number of commits per author in a rolling window based on
+#' the git log.
+#'
+#' @param project_git_log a parsed git log obtained from \code{\link{parse_gitlog}}
```
jseto808, line 190: Kaiaulu functions name this parameter as `project_git`.

Similar naming can be found:
https://github.com/sailuh/kaiaulu/blob/e88975dc028b3efd8ba6c293836227df8feb00cd/R/metric.R#L97

R/metric.R
```diff
@@ -181,6 +181,115 @@ commit_message_id_coverage <- function(git_log,commit_message_id_regex){
...
+  }, by = .(author_name_email)]
+
+  return(result[])
+}
+
+#' Productivity Author Churn Metric
+#'
+#' Calculates the churn per author in a rolling window based on the git log.
+#'
+#' @param project_git_log a parsed git log obtained from \code{\link{parse_gitlog}}
```
jseto808, line 244: Same issue as above: naming of `project_git`

jseto808 (changes requested):
The input parameter is named `project_git_log` in both new functions, while existing metric functions use either `project_git` or `git_log`. This should be made consistent with one of the two established conventions. Additionally, the notebook introduction references `jira_developer_communication_count` as one of the functions showcased but the function does not exist in `metric.R` and is not demonstrated anywhere in the notebook. Not sure what the reference is for, but if the function is not implemented, it should be removed.

Otherwise, the notebook was knitted successfully.

usradam:
Code has been moved to PR https://github.com/sailuh/kaiaulu/pull/403.

